PR #7, opened 2016-11-20
[API-docs] Update Event/Request Fields

haganbmj:
Thanks for picking up work on this for OBS Studio. Fiddled around with it this evening and wanted to update the Protocol doc with some of the expected fields.

I couldn't seem to find a way to trigger [RecordingStopping](https://github.com/Palakis/obs-websocket/blob/master/WSEvents.cpp#L154), so I left a "Not Yet Implemented" in the doc.

Also, I have no idea what preview-only means. I tried disabling my preview and projecting to another monitor and still couldn't get that flag to change.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -1,70 +1,73 @@
 obs-websocket protocol reference
 ================================
 
-## General introduction
+## General Introduction
 **This document is still a WIP. Some things are missing (but won't stay like this for long).**
 
-Messages exchanged between the client and the server are JSON objects.
-The protocol in general in based on the OBS Remote protoctol created by Bill Hamilton, with new commands specific to OBS Studio. 
+Messages exchanged between the client and the server are JSON objects.  
+The protocol in general is based on the OBS Remote protocol created by Bill Hamilton, with new commands specific to OBS Studio.
 
 ## Events
 
 ### Description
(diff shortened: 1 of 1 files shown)


Palakis:
Thanks @haganbmj !
Regarding the `preview-only` field, it is always false and is there to keep some sort of compatibility with the OBSRemote protocol.
Eventually, when the first clients and libraries will show up and the protocol evolves, this field might get removed.

haganbmj:
Gotcha, that makes perfect sense.
Seems to work well so far that I've encountered; quick and easy to consume.

Palakis:
Copy-pasting when tired can lead to mistakes... One of the event types was handled twice in the event handler.   
Since https://github.com/Palakis/obs-websocket/commit/d1a142cefd3288c99827da8854085d673d90d82b, `RecordingStopping` should be working.

haganbmj:
Ahh, I missed that typo too. Nice catch.

haganbmj:
Got the basic JavaScript consumption of this done yesterday and created a repo for that. I never used the old obsremote, but looking at the source code it looks like it wouldn't be difficult to create a new client.obsremote.com to match the api changes. I'll probably get a start on that tonight or tomorrow.

PR #40, opened 2017-02-22
Add german translation

Frahmer:
Add german translation for menu entry and settings dialog.


Palakis:
Sehr gut !

Palakis:
@Frahmer Two new strings have been added to the UI (see [here](https://github.com/Palakis/obs-websocket/blob/master/data/locale/en-US.ini)). Can you also add these to the German translation please? (Many thanks in advance)

PR #41, opened 2017-02-23
Add german translation

Frahmer:
Add "ServerEnable" and "ServerPort" translation.


Palakis:
Awesome! Thank you very much.

PR #45, opened 2017-02-26
Toggle source on specified scene

mikhailswift:
Adds the ability to specify a scene when calling SetSourceRender.


Palakis:
Looks good to me. Merging right now.

PR #87, opened 2017-06-13
Set Text for text_gdiplus

pantonvich:


Palakis:
This can't be accepted as is. The code for `SetSceneItemRender` isn't the place for source property changes.

I can advise you two different ways to achieve what you'd like to achieve:
- Implement `SetGDIText`/`GetGDIText` request types to get/set the text value or settings on a GDI+ text source.
- Implement more generic `SetSourceProperties`/`GetSourceProperties` request types, which can also manipulate Browser Sources and others

pantonvich:
hadn't really thought about needing a get so that's a good idea - 
I've got the other settings working - font-size, bk-color, bk-opacity and color working.
I'll look at moving what I have to SetGDIText and creating a GetGDIText
If there are other settings that would be useful people can chime in and I'll see about getting them added.

I tried to get browser source complied as I was thinking getting to the url setting would work the same as the gdi text - but have been unable due to an issue with pre-complied CEF - after mult attempts over two days I've given up on browser source... 

I'm not very good with git - and common practices so should I close this pull and reopen a new one once I get the Set and Get ready?

pantonvich:
I'm so sorry I'm making a mess with github - I'm really not sure what I need to do to fix the github mess I've made here - again I'm so sorry.

pantonvich:
I'm going to close 91 and 92 and  try again

PR #98, opened 2017-06-30
General: Toggle to log incoming requests

haganbmj:
This might not be desired.
Adds a toggle to the settings dialogue to enabled/disable logging incoming requests in their original json format.

```
21:00:30.658: [obs-websocket] >> '{"scene-name":"Other","request-type":"SetCurrentScene","message-id":"3"}'
```


Palakis:
Everything's fine overall. What about responses?

haganbmj:
Sure, that sounds pretty reasonable. I'll get that updated later today.

haganbmj:
That was not ready to merge.

haganbmj:
Well it might be? Guess we'll find out.

Palakis:
@haganbmj What do you have in mind?

haganbmj:
That commit is broken, looks like I'm emitting a pointer instead of the json data.

Palakis:
Seems fine to me. Passing a `const char*` as a `%s` to `blog()` will make it recognized as a string by the latter.

haganbmj:
Yeah, but passing that to `_srv->broadcast` doesn't appear to.

haganbmj:
Also getting issues with `_client->sendTextMessage(json)`.
first message with both are fine, but subsequent messages just get mangled.

I don't have a build environment set up on this machine, so I was just using appveyor to build for me over on my fork.

Palakis:
I refactored some parts. Once it's built you can test it.
Commit: https://github.com/Palakis/obs-websocket/commit/54a16f4d2f7bddcb16265478f0d7044c05242c8b and https://github.com/Palakis/obs-websocket/commit/acffacd67db1a73a1a07d12edeed1439a7669351
On Travis: https://travis-ci.org/Palakis/obs-websocket/builds/250102188
On Appveyor: https://ci.appveyor.com/project/Palakis/obs-websocket/build/master-227

Palakis:
I believe you called `obs_data_release` too soon, in case the pointer belongs to the `obs_data_t*` context too.

Palakis:
Ah, maybe what I mentioned above is the cause.

Code for `obs_data_get_json` (taken from [here](https://github.com/jp9000/obs-studio/blob/2d84da2f7e38c6d26acfe2d3a1d634d6dc360eae/libobs/obs-data.c#L725-L738)):
```cpp
const char *obs_data_get_json(obs_data_t *data)
{
	if (!data) return NULL;

	/* NOTE: don't use libobs bfree for json text */
	free(data->json);
	data->json = NULL;

	json_t *root = obs_data_to_json(data);
	data->json = json_dumps(root, JSON_PRESERVE_ORDER | JSON_INDENT(4));
	json_decref(root);

	return data->json;
}
```
Note the comment above `free(data->json);`.

haganbmj:
Ah that's possible.
I'll take a look with what appveyor spits out.

haganbmj:
In hindsight using both a settings toggle and requiring `--verbose` doesn't make a whole lot of sense. One of those should go. Either the logs should be using `LOG_INFO` or there shouldn't be a settings option and it should attempt to `LOG_DEBUG` everything.

haganbmj:
Rearranging those does appear to have worked by the way.
Threw a number of requests at it and didn't run into any issues.

PR #100, opened 2017-07-06
Custom stream settings

yinzara:
I took some of your ideas from your rtmp-settings branch and combined them with my prior changes.  Your access of the service configuration via the output stream was actually not required (see the code changes),

I decided to simplify some of the names of things and just go with "StreamSettings" as the term used in the protocol in the tab inside OBS Studio settings is called "Stream".

I have added GetStreamSettings, SetStreamSettings, and SaveStreamSettings requests (the "SaveStreamSettings" can also be achieved with a  'save' request parameter on the SetStreamSettings call).  This allows you to (on disk or only in memory) change the streaming server settings. SetStreamSettings takes in the same structure as the response from GetStreamSettings and also supports "patch" like operations as it allows you to only specify some settings (as long as you don't change the 'type') and the rest get taken from the current streaming server settings.  The primary use of these methods would be to support a configuration dialog remotely. Unfortunately there is still no way of getting the possible settings for the rtmp_common type to render a dialog remotely (i.e. what's in the services.json file in obs-studio), but maybe we can add them in a future revision.

I also opted to enhance StartStreaming and StartStopStreaming to support a new 'stream' (similar to 'with_settings' in your branch) parameter that has the exact same structure as the request from SetStreamSettings.  The 'stream' parameter also supports a "metadata" parameter as well that enhances the query string of the 'key' passed to the streaming server to allow you to specify additional metadata about the stream (I even dealt with percent encoding in case your values include bad characters).  Servers like Red5 allow you to read these as "parameters" of the stream.

This should make your rtmp-settings branch obsolete as I've incorporated everything you were doing there into this pull request.

### DIFF
--- a/Utils.cpp
+++ b/Utils.cpp
@@ -20,6 +20,7 @@ with this program. If not, see <https://www.gnu.org/licenses/>
 #include <obs.hpp>
 #include <QMainWindow>
 #include <QDir>
+#include <QUrl>
 #include "Utils.h"
 #include "obs-websocket.h"
 
@@ -464,4 +465,56 @@ bool Utils::SetRecordingFolder(const char* path)
 
 	config_save(profile);
 	return true;
+}
+
+QString* Utils::ParseDataToQueryString(obs_data_t * data) {
+	QString * query = nullptr;
+	if (data) {
+		obs_data_item_t * item = obs_data_first(data);
+		if (item) {
+			query = new QString();
+			bool isFirst = true;
+			do {
+				if (!obs_data_item_has_user_value(item)) {
+					continue;
+				}
+				if (!isFirst) {
+					query->append('&');
+				} else {
+					isFirst = false;
+				}
+				const char * attrName = obs_data_item_get_name(item);
+				query->append(attrName).append("=");
+				switch (obs_data_item_gettype(item)) {
+					case OBS_DATA_BOOLEAN:
+						query->append(obs_data_item_get_bool(item)?"true":"false");
+						break;
+					case OBS_DATA_NUMBER:
+						switch (obs_data_item_numtype(item)) {
+							case OBS_DATA_NUM_DOUBLE: {
+								double value = obs_data_item_get_double(item);
+								query->append(QString::number(value));
+								break;
+							}
+							case OBS_DATA_NUM_INT: {
+								long long value = obs_data_item_get_int(item);
+								query->append(QString::number(value));
+								break;
+							}
+							case OBS_DATA_NUM_INVALID:
+								break;
+						}
+						break;
+					case OBS_DATA_STRING:
+						query->append(QUrl::toPercentEncoding(QString(obs_data_item_get_string(item))));
+						break;
+					default:
+						//other types are not supported
+						break;
+				}
+			} while ( obs_data_item_next( &item ) );
+		}
+	}
+	
+	return query;
 }
--- a/Utils.cpp
+++ b/Utils.cpp
@@ -467,40 +467,42 @@ bool Utils::SetRecordingFolder(const char* path)
 	return true;
 }
 
-QString* Utils::ParseDataToQueryString(obs_data_t * data) {
-	QString * query = nullptr;
-	if (data) {
-		obs_data_item_t * item = obs_data_first(data);
-		if (item) {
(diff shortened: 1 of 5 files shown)


Palakis:
> Your access of the service configuration via the output stream was actually not required (see the code changes),

This was done that way because `rtmp-settings` was a hack for when OBS' Frontend API had no method to get/set/save streaming settings. At the same time I had submitted a PR to add these to the frontend API.

Palakis:
> Unfortunately there is still no way of getting the possible settings for the rtmp_common type to render a dialog remotely (i.e. what's in the services.json file in obs-studio), but maybe we can add them in a future revision.

https://github.com/Palakis/obs-websocket/blob/rtmp-settings/WSRequestHandler.cpp#L922-L978

Utils.cpp
```diff
@@ -464,4 +465,56 @@ bool Utils::SetRecordingFolder(const char* path)
...
+QString* Utils::ParseDataToQueryString(obs_data_t * data) {
+	QString * query = nullptr;
+	if (data) {
+		obs_data_item_t * item = obs_data_first(data);
+		if (item) {
+			query = new QString();
+			bool isFirst = true;
+			do {
+				if (!obs_data_item_has_user_value(item)) {
+					continue;
```
Palakis: It is advised to not put brackets if a code block has only one line.
```cpp
if (!obs_data_item_has_user_value(item))
    continue;
```

Utils.cpp
```diff
@@ -464,4 +465,56 @@ bool Utils::SetRecordingFolder(const char* path)
 
 	config_save(profile);
 	return true;
-}
\ No newline at end of file
+}
+
+QString* Utils::ParseDataToQueryString(obs_data_t * data) {
```
Palakis: Opening brackets on a new line, both for function and conditionals.

```cpp
QString* Utils::ParseDataToQueryString(obs_data_t* data)
{
```

```cpp
if(data)
{
    obs_data_item_t* item = obs_data_first(data);
```

Utils.cpp
```diff
@@ -464,4 +465,56 @@ bool Utils::SetRecordingFolder(const char* path)
...
+	QString * query = nullptr;
+	if (data) {
+		obs_data_item_t * item = obs_data_first(data);
```
Palakis: Align pointer symbol with the type, not with the variable name or inbetween.

```cpp
obs_data_item_t* item = obs_data_first(data);
```

WSRequestHandler.h
```diff
@@ -33,8 +34,9 @@ class WSRequestHandler : public QObject
 		~WSRequestHandler();
 		void processIncomingMessage(QString textMessage);
 		bool hasField(const char* name);
-
+
```
Palakis: Ghost indentation.

WSRequestHandler.cpp
```diff
@@ -401,8 +413,90 @@ void WSRequestHandler::HandleStartStreaming(WSRequestHandler* req)
 {
 	if (obs_frontend_streaming_active() == false)
 	{
+		obs_data_t* streamData = obs_data_get_obj(req->data, "stream");
```
Palakis: In `rtmp-settings`, that behaviour was kinda mandatory because there wasn't an other way to set streaming settings before starting it. Now that we have the proper methods in the frontend API, there's not need to implement that behavior. A user willing to set streaming settings should call `SetStreamSettings` followed by `StartStreaming`.

Palakis:
Btw, thanks a lot for these additions! This is a very awaited feature.

WSRequestHandler.cpp
yinzara: Interestingly enough there is actually a valid use case for this.  Specifically for the case of passing metadata.  Usually metadata is specific to a stream and if that stream was stopped you would not want that metadata to pollute future streams.  If you just call SetStreamSettings, you have the risk of doing that if the stop stream event isn't received (and I found some scenarios in which it was). That's why I primarily wrote this functionality.

yinzara: It also allows you to have a separate configuration specifically just for the web socket client so that UI users would not see the changes to the settings and that if they started streaming they would remain with the original streaming server settings.

Palakis: I see. Ok, then.

Issue #90, opened 2017-06-16
Feature Request: Save replay buffer


XBase-xOR:
#### Issue type
Feature request

#### Description
I have written a tool that does an instant replay while live. Right now I have defined the F14 hotkey (which doesn't exist on my keyboard so it's otherwise not used) in OBS for the "Save Replay Buffer" action and then my tool sends an F14 keystroke to the OBS process using PostMessage().
Although that is working it's still somewhat hackish. Since the tool is already connected to OBS using your fine websocket interface it would be great if I could just use the websocket interface for that.

EDIT for further clarification:
Check [this screenshot](https://x-base.org/img/OBS_replay_buffer.png). The hotkeys with a green border are for functions that can also be triggered from the GUI on the main window. The hotkey with a red border however is the only way I know of to trigger writing of the file, there is no GUI button equivalent. And this is the functionality I am asking to be able to trigger from the websocket interface.

#### Technical information
- **Operating System** : Windows 10 64 Bit with Creators Update
- **OBS Studio version** : 18.0.1

Palakis:
It's in the works: https://github.com/Palakis/obs-websocket/pull/104

TStod:
Just found this repo. Noticed replay events/requests were not implemented. Went to file a feature request. Find this.

FeelsAmazingMan

XBase-xOR:
@Palakis Thanks so much dude, changed my tool from using hotkey to the new API method and it works like a charm!

PR #104, opened 2017-07-11
Add Replay Buffer Events and Requests

lsdaniel:
Documentation is included, not really much, but I needed events for the replay buffer starting and stopping. These events have only been available as of 17.0.0 (I think). I don't think that the OBS API has an event for saving the replay yet, so that can't happen yet :/.


lsdaniel:
I have found requests to start/stop the replay buffer, but I haven't found one to flush the replay buffer to disk. Anyone have any ideas?

Palakis:
Thanks for this, that is good quality code. I've fixed conflicts caused by my recent refactor of the codebase, along with some style nitpicks.

> I have found requests to start/stop the replay buffer, but I haven't found one to flush the replay buffer to disk. Anyone have any ideas?

I'll have a look at OBS' API and come back to you.

Palakis:
Looks like even OBS itself doesn't have anything to flush a replay buffer. Maybe deleting or emptying the file will do the trick. Is it what you want to achieve, @lsdaniel?

lsdaniel:
@Palakis Sorry for wording it that way, I meant save the current replay buffer (what is in memory) to the disk. This is normally achieved by using a hotkey, but there is no front-end function to do it using the libobs api. Unless I am wrong, which I may be. I filed a bug on [OBS's bug tracker](https://obsproject.com/mantis/view.php?id=955), but have heard nothing so far. The farthest I've understood how the replay buffer works is by looking at [this function](https://github.com/jp9000/obs-studio/blob/cfe77fdaaeacb78027e8c045739ead7a6f673302/plugins/obs-ffmpeg/obs-ffmpeg-mux.c#L685) in the obs code.

Palakis:
AFAIK the replay buffer is constantly being written to while active, so no need to trigger a manual save or something.

Palakis:
The replay buffer actually being a file.

XBase-xOR:
OBS doesn't have any GUI element that can trigger writing the file. Pressing a hotkey that is configured [here](https://x-base.org/img/OBS_replay_buffer.png) (red border) is the only way.

The replay buffer in memory is constantly updated in memory and not written to a file unless you trigger to write it to a file by using that hotkey.

I also updated #90 for further clarification.

Palakis:
Thanks @XBase-xOR. I'm not a regular user of this feature so I didn't have all the details.
I can't find this "Save replay buffer" hotkey, both in 19.0.3 and the latest OBS master codebase. Is there something special to enable?

Palakis:
@lsdaniel I managed to manually trigger buffer saves through the hotkey system!

Palakis:
@XBase-xOR nevermind, If found the option to enable the Replay buffer and the hotkey now shows up.

XBase-xOR:
Ah OK, I didn't even know that OBS hides it if the replay buffer is disabled.

Palakis:
Things left to do:
- Return an error if the replay buffer is disabled in OBS' settings
- Handle the case where no hotkey is bound to save the replay buffer to disk: bind a dummy hotkey? return an error?

TStod:
It is probably worth noting that the replay buffer is only enabled when using the `Simple` output mode with the `Enable Replay Buffer` checkbox selected.

Hotkey settings for capturing/enabling/disabling replay are stored but hidden if the above settings are not set/enabled.

WSRequestHandler.cpp
```diff
@@ -1638,7 +1736,7 @@ void WSRequestHandler::HandleSetRecordingFolder(WSRequestHandler* req) {
 /**
  * Get the path of  the current recording folder.
  *
- * @return {Stsring} `rec-folder` Path of the recording folder.
+ * @return {String} `rec-folder` Path of the recording folder.
```
haganbmj: Thanks for fixing these. Sloppy typos.

Palakis: Shit happens 😄

XBase-xOR:
@Palakis Thanks so much dude, changed my tool from using hotkey to the new API method and it works like a charm!

ihrishix:
@Palakis @haganbmj Is there any way i can remotely enable the Replay Buffer from obs settings and set the duration ? (Obs settings > output > replay buffer)

Issue #44, opened 2017-02-26
Machine-readable protocol definition


Palakis:
As of now, PROTOCOL.md is written by hand. 
The idea would be to have protocol definitions in a machine-readable format somewhere in the repo (a separate file or inline comments in the source code) that will serve as a base for an automatically generated PROTOCOL.md and/or an eventual test suite.

haganbmj:
Doxygen is the most popular in C++, right?

PR #106, opened 2017-07-16
Docs: Initial generation of docs from code comments

haganbmj:
This is unfinished, but I wanted to at least open up a PR as a proof of concept and to see if this is worth finishing.

Reads from source files and generates the protocol document from that using a mustache template.

Proposed TODOs:
- Document the rest of the requests/events.
- ~~Child properties of response json objects (such as in the case of Stream settings).~~
- ~~Travis script to generate docs and publish them to gh-pages.~~ (edit: publish back to branch)
- Better categorization and sorting of request/event types.
- General code cleanup (which would probably involve dumping this into its own folder too).

Related to #44


docs/comments.json
```diff
@@ -0,0 +1,754 @@
+{
```
haganbmj: This file is generated from the code comments.

docs/protocol.md
```diff
@@ -0,0 +1,293 @@
+<!DOCTYPE md>
```
haganbmj: This file is generated based on protocol.mustache and comments.json

WSRequestHandler.cpp
```diff
@@ -1,21 +1,21 @@
...
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-GNU General Public License for more details.
-
-You should have received a copy of the GNU General Public License along
-with this program. If not, see <https://www.gnu.org/licenses/>
-*/
+/**
+ * obs-websocket
+ * Copyright (C) 2016-2017	Stéphane Lepin <stephane.lepin@gmail.com>
+ * Copyright (C) 2017	Brendan Hagan <https://github.com/haganbmj>
```
haganbmj: Whoops, didn't mean to do that. Blindly copied this over from the other file. I'll revert when I have a chance.

Palakis:
Thanks a lot for this! That last commit is me fixing the mess caused in PRs caused by my recent code style refactor.

haganbmj:
No worries, do you have a outline of the code style you want to follow?

Palakis:
The only two guidelines for code style are:
- Function blocks and conditionals: don't dedicate a line for a block's opening bracket (it was the case before, but I changed my mind)
- Variables: Align the `*` pointer symbol with the type, not the name

Palakis:
> General code cleanup (which would probably involve dumping this into its own folder too).

Did you see anything shocking or out of place? :smile:

Palakis:
@haganbmj I'd like to help you on this. Have you enabled the "Allow edits from maintainers" option?

haganbmj:
I believe so, you were able to merge to it. I'm on mobile now, I can check later. 
I think I'll try to locate all of this into just a docs folder so it's not as much clutter.

haganbmj:
Yeah maintainer edits are enabled.

haganbmj:
Alright, tidied this up a bit.
Rebased on latest master, moved all the docs related stuff into the `docs` directory.
Split the js into two files and added a brief readme.

What I'd love is to get the rest of the methods/events commented and a travis script that generates docs on every commit, I'm thinking it might be best to just push changes back to master since that's where you're releasing from.

- Commit, travis generates docs after a build.
- If differences exist, push the doc changes back to master.
- Some logic to avoid wasting build time on travis commits maybe.

Mustache is a bit limiting due to how it needs the object to be structured (only being able to iterate arrays is killer), so I might see if there's any alternative or some simple way to overcome that.

haganbmj:
Alright, so this will generate docs and push them back to master assuming you have a `GH_TOKEN` in your travis config.
I'm running into some issues with line endings though. It definitely appears to be ignoring some line endings during the build on OSX which is causing some differences in the output there.

My only other issue is that whenever this commits back to master it triggers a full rebuild in both travis and appveyor - not sure if that's really desirable since the protocol doc isn't included in a deployment.

haganbmj:
Eh, "fixed" the line ending issue by just stripping out line numbers and such from the generated files.
Split the documentation build off into its own travis job.
Not sure why the OSX builds are having trouble getting started, probably a travis issue. It was happening all evening.

haganbmj:
osx builds are hanging again.
Switched to handlebars. Cleaner templating. 
All that's really left is to document everything.

haganbmj:
Everything's documented.
I'll rebase after #109 gets merged and update the docs for that, then this should be ready to go.

Palakis:
> I'm thinking it might be best to just push changes back to master since that's where you're releasing from.

> My only other issue is that whenever this commits back to master it triggers a full rebuild in both travis and appveyor - not sure if that's really desirable since the protocol doc isn't included in a deployment.

I think it would be better to have documentation on GitHub Pages. It will avoid triggering a new build if I add exclusions on Travis and AppVeyor for the GH Pages branch. But I'll take care of this on my end, you've already done amazing work!

haganbmj:
Yeah no worries. I'll get to it in 12 hours or so.

haganbmj:
So the alternative is that I just have travis ignore its own commits, which I think is a cleaner solution in this case. 
I'd rather leave the Protocol in master in this case so that you can have your tags point to both the source files and the documentation? Adding an entire branch for just one file seems unnecessary.

Rebased, added the SyncOffset documentation, updated Travis to not trigger on doc commits.
I squashed commits, but I'd still recommend either Squash and Merge or Rebase and Merge instead of generating a Merge Commit.

I also purposefully didn't include the updated Protocol with SyncOffset so that it would trigger a travis build. Double check that you have `GH_TOKEN` as an environment variable in travis.

CI/generate-docs.sh
```diff
@@ -0,0 +1,32 @@
...
+    echo "-- Skipping documentation deployment because this is either a pull request or a non-master branch."
+    exit 0
+fi
+
+REMOTE_URL="$(git config remote.origin.url)"
+TARGET_REPO=${REMOTE_URL/https:\/\/github.com\//github.com/}
+GITHUB_REPO=https://${GH_TOKEN:-git}@${TARGET_REPO}
+
+git config user.name "Travis CI"
+git config user.email "$COMMIT_AUTHOR_EMAIL"
```
Palakis: Do I have to add this environment variable to Travis as well?

haganbmj: This is pulled from the commit that started the build.

PR #109, opened 2017-08-07
Add request types: SetSyncOffset and GetSyncOffset

dragonbane0:
Adds those 2 request types for the functions "obs_source_set_sync_offset" and "obs_source_get_sync_offset". We plan to gracefully auto adjust an audio source delay to match a twitch stream (based on its delay). So this would be great to have!

### DIFF
--- a/WSRequestHandler.cpp
+++ b/WSRequestHandler.cpp
@@ -77,8 +77,8 @@ WSRequestHandler::WSRequestHandler(QWebSocket* client) :
     messageMap["ToggleMute"] = WSRequestHandler::HandleToggleMute;
     messageMap["SetMute"] = WSRequestHandler::HandleSetMute;
     messageMap["GetMute"] = WSRequestHandler::HandleGetMute;
-	messageMap["SetSyncOffset"] = WSRequestHandler::HandleSetSyncOffset;
-	messageMap["GetSyncOffset"] = WSRequestHandler::HandleGetSyncOffset;
+    messageMap["SetSyncOffset"] = WSRequestHandler::HandleSetSyncOffset;
+    messageMap["GetSyncOffset"] = WSRequestHandler::HandleGetSyncOffset;
     messageMap["GetSpecialSources"] = WSRequestHandler::HandleGetSpecialSources;
 
     messageMap["SetCurrentSceneCollection"] = WSRequestHandler::HandleSetCurrentSceneCollection;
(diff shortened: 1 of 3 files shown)


WSRequestHandler.cpp
```diff
@@ -710,6 +712,58 @@ void WSRequestHandler::HandleGetMute(WSRequestHandler* req) {
...
+	!req->hasField("offset")) {
+	req->SendErrorResponse("missing request parameters");
+	return;
+    }
+
+    const char* source_name = obs_data_get_string(req->data, "source");
+    int64_t source_sync_offset = (int64_t)obs_data_get_int(req->data, "offset");
+
+    if (source_name == NULL || strlen(source_name) < 1 ||
+	source_sync_offset < 0 {
```
haganbmj: Build appears to be failing due to a missing parenthesis here.

Palakis:
@dragonbane0 Thanks for this contribution! Make sure to use 4-spaces indents instead of tabs, though.

WSRequestHandler.cpp
```diff
@@ -713,55 +713,55 @@ void WSRequestHandler::HandleGetMute(WSRequestHandler* req) {
 }
 
 void WSRequestHandler::HandleSetSyncOffset(WSRequestHandler* req) {
-	if (!req->hasField("source") ||
-		!req->hasField("offset")) {
-		req->SendErrorResponse("missing request parameters");
-		return;
-	}
+    if (!req->hasField("source") ||
+	!req->hasField("offset")) {
```
Palakis: Make sure to use 4-spaces indents all the way.

WSRequestHandler.cpp
```diff
@@ -710,6 +712,58 @@ void WSRequestHandler::HandleGetMute(WSRequestHandler* req) {
...
+	req->SendErrorResponse("missing request parameters");
+	return;
+    }
+
+    const char* source_name = obs_data_get_string(req->data, "source");
+    int64_t source_sync_offset = (int64_t)obs_data_get_int(req->data, "offset");
+
+    if (source_name == NULL || strlen(source_name) < 1 ||
```
Palakis: It's C++11 code, so use `nullptr` instead of `NULL`. Or better: use `!source_name` as the condition instead, like below with `!source`.

dragonbane0:
Fixed the if statement and added the new request types to the documentation

haganbmj:
Cool, looks to be in order.
Just now recognized your name from zelda runs dragonbane. Nice work.

dragonbane0:
I tested this out and it works pretty much perfectly. We have an embedded twitch player in our browser source where we use js and the provided API to check the delay to host every 10 seconds and if it changes we slowly adjust the sync offset of our audio source where we have the streamer of the re-streamed channel commentating his gameplay until it matches the video delay. A 50 ms change every second or so is basically unnoticeable even while someone is speaking non stop so it only takes around 2.5 minutes to get from 0 to the average 8 second delay on twitch. Very neat

PR #117, opened 2017-08-25
docs: Add @since

haganbmj:
Just started on this, but I'm boarding a plane in 5 minutes. I'll finish up the rest of the requests later and clean up the generated doc a bit to account for unreleased stuff more gracefully.


haganbmj:
Alright, got them all documented. I used @since unreleased for things that haven't been released yet. I guess those'll need to be replaced with whatever version is next.

PR #136, opened 2017-10-04
[Safe For Merge] Unified Setter and Getter for scene specific source properties

TStod:
See line specific questions below @Palakis @haganbmj

### DIFF
--- a/WSRequestHandler.cpp
+++ b/WSRequestHandler.cpp
@@ -1617,7 +1617,6 @@ void WSRequestHandler::HandleSetSceneItemProperties(WSRequestHandler* req) {
             // Send an error in the else statement?
             // Append an error message to the response?
         }
-        obs_sceneitem_set_pos(scene_item, &new_position);
     }
 
     obs_sceneitem_release(scene_item);
--- a/WSRequestHandler.cpp
+++ b/WSRequestHandler.cpp
@@ -56,8 +56,7 @@ WSRequestHandler::WSRequestHandler(QWebSocket* client) :
     messageMap["SetSceneItemPosition"] = WSRequestHandler::HandleSetSceneItemPosition;
     messageMap["SetSceneItemTransform"] = WSRequestHandler::HandleSetSceneItemTransform;
     messageMap["SetSceneItemCrop"] = WSRequestHandler::HandleSetSceneItemCrop;
-    // This method name needs work. Perhaps just "GetSceneItemProperties"
-    messageMap["GetSceneItemSceneProperties"] = WSRequestHandler::HandleGetSceneItemSceneProperties;
+    messageMap["GetSceneItemProperties"] = WSRequestHandler::HandleGetSceneItemProperties;
     messageMap["ResetSceneItem"] = WSRequestHandler::HandleResetSceneItem;
 
     messageMap["GetStreamingStatus"] = WSRequestHandler::HandleGetStreamingStatus;
@@ -1350,7 +1349,23 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
  * @param {String (optional)} `scene-name` the name of the scene that the source item belongs to. Defaults to the current scene.
  * @param {String} `item` The name of the source.
  *
- * @return TO POPULATE
+ * @return {String} `name` The name of the source.
+ * @return {int} `position.x` The x position of the source from the left.
+ * @return {int} `position.y` The y position of the source from the top.
+ * @return {double} `rotation` The clockwise rotation of the item in degrees around the point of alignment.
+ * @return {int?string?} `alignment` The point on the source that the item is manipulated from.
+ * @return {double} `scale.x` The x-scale factor of the source.
+ * @return {double} `scale.y` The y-scale factor of the source.
+ * @return {int} `crop.top` The number of pixels cropped off the top of the source before scaling.
+ * @return {int} `crop.right` The number of pixels cropped off the right of the source before scaling.
+ * @return {int} `crop.bottom` The number of pixels cropped off the bottom of the source before scaling.
+ * @return {int} `crop.left` The number of pixels cropped off the left of the source before scaling.
+ * @return {bool} `visible` If the source is visible.
+ * @return {bool/Object} `bounds` False if bounds are not turned on. Object if bounds are turned on.
+ * @return {String} `bounds.type` Type of bounding box.
+ * @return {int?string?} `bounds.alignment` Alignment of the bounding box.
+ * @return {double} `bounds.width` Width of the bounding box.
+ * @return {double} `bounds.height` Height of the bounding box.
  *
  * @api requests
  * @name GetSceneItemSceneProperties
@@ -1383,7 +1398,6 @@ void WSRequestHandler::HandleGetSceneItemSceneProperties(WSRequestHandler* req) 
         // is name even needed here?
         obs_data_set_string(data, "name", item_name);
         
-        // is nested ideal?
         obs_data_t* pos_data = obs_data_create();
         vec2 pos;
         obs_sceneitem_get_pos(scene_item, &pos);
@@ -1393,7 +1407,7 @@ void WSRequestHandler::HandleGetSceneItemSceneProperties(WSRequestHandler* req) 
 
         obs_data_set_double(data, "rotation", obs_sceneitem_get_rot(scene_item));
 
-        // investigate this and maybe convert so string in switch statment
+        // investigate this and maybe convert so string in switch statement
         // #define OBS_ALIGN_CENTER (0)
         // #define OBS_ALIGN_LEFT   (1<<0)
         // #define OBS_ALIGN_RIGHT  (1<<1)
@@ -1461,8 +1475,8 @@ void WSRequestHandler::HandleGetSceneItemSceneProperties(WSRequestHandler* req) 
             obs_data_set_int(bounds_data, obs_sceneitem_get_bounds_alignment(scene_item));
             vec2 bounds;
             obs_sceneitem_get_bounds(scene_item, &bounds);
-            obs_data_set_float(bounds_data, "width", bounds.x);
-            obs_data_set_float(bounds_data, "height", bounds.y);
+            obs_data_set_double(bounds_data, "width", bounds.x);
+            obs_data_set_double(bounds_data, "height", bounds.y);
             obs_data_set_obj(data, "bounds", bounds_data);
         }
 
--- a/WSRequestHandler.cpp
+++ b/WSRequestHandler.cpp
@@ -1543,10 +1543,10 @@ void WSRequestHandler::HandleSetSceneItemProperties(WSRequestHandler* req) {
         obs_sceneitem_get_pos(scene_item, &old_position);
         obs_data_t* req_position = obs_data_get_obj(req->data, "position");
         vec2 new_position = old_position;
-        if (req_crop->hasField("x")) {
+        if (obs_data_has_user_value(req_position, "x")) {
             new_position.x = obs_data_get_int(req_position, "x");
         }
-        if (req_crop->hasField("y")) {
+        if (obs_data_has_user_value(req_position, "y")) {
             new_position.y = obs_data_get_int(req_position, "y");
         }
         obs_sceneitem_set_pos(scene_item, &new_position);
@@ -1557,16 +1557,16 @@ void WSRequestHandler::HandleSetSceneItemProperties(WSRequestHandler* req) {
         obs_sceneitem_get_crop(scene_item, &old_crop);
         obs_data_t* req_crop = obs_data_get_obj(req->data, "crop");
         obs_sceneitem_crop new_crop = old_crop;
-        if (req_crop->hasField("top")) {
+        if (obs_data_has_user_value(req_crop, "top")) {
             new_crop.top = obs_data_get_int(req_crop, "top");
         }
-        if (req_crop->hasField("right")) {
+        if (obs_data_has_user_value(req_crop, "right")) {
             new_crop.right = obs_data_get_int(req_crop, "right");
         }
-        if (req_crop->hasField("bottom")) {
+        if (obs_data_has_user_value(req_crop, "bottom")) {
             new_crop.bottom = obs_data_get_int(req_crop, "bottom");
         }
-        if (req_crop->hasField("left")) {
+        if (obs_data_has_user_value(req_crop, "left")) {
             new_crop.left = obs_data_get_int(req_crop, "left");
         }
         obs_sceneitem_set_crop(scene_item, &new_crop);
--- a/WSRequestHandler.cpp
+++ b/WSRequestHandler.cpp
@@ -1362,7 +1362,6 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
  * @return {int} `crop.bottom` The number of pixels cropped off the bottom of the source before scaling.
  * @return {int} `crop.left` The number of pixels cropped off the left of the source before scaling.
  * @return {bool} `visible` If the source is visible.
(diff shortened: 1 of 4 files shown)


WSRequestHandler.cpp
```diff
@@ -56,6 +56,8 @@ WSRequestHandler::WSRequestHandler(QWebSocket* client) :
     messageMap["SetSceneItemPosition"] = WSRequestHandler::HandleSetSceneItemPosition;
     messageMap["SetSceneItemTransform"] = WSRequestHandler::HandleSetSceneItemTransform;
     messageMap["SetSceneItemCrop"] = WSRequestHandler::HandleSetSceneItemCrop;
+    // This method name needs work. Perhaps just "GetSceneItemProperties"
+    messageMap["GetSceneItemSceneProperties"] = WSRequestHandler::HandleGetSceneItemSceneProperties;
```
TStod: Method name?

WSRequestHandler.cpp
```diff
@@ -1343,6 +1345,141 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
 }
 
 /**
+ * Gets the scene specific properties of the specified source item.
+ *
+ * @param {String (optional)} `scene-name` the name of the scene that the source item belongs to. Defaults to the current scene.
+ * @param {String} `item` The name of the source.
+ *
+ * @return TO POPULATE
```
TStod: If nested properties how do I properly show the return types?

WSRequestHandler.cpp
```diff
@@ -1343,6 +1345,141 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+        req->SendErrorResponse("requested scene doesn't exist");
+        return;
+    }
+
+    obs_sceneitem_t* scene_item = Utils::GetSceneItemFromName(scene, item_name);
+    if (scene_item) {
+        obs_data_t* data = obs_data_create();
+
+        // is name even needed here?
+        obs_data_set_string(data, "name", item_name);
```
TStod: Does it make sense to keep this in this method?

WSRequestHandler.cpp
```diff
@@ -1343,6 +1345,141 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+        
+        // is nested ideal?
+        obs_data_t* pos_data = obs_data_create();
```
TStod: I use nested objects for position, crop, scale, and bounds. Do we prefer a flat implementation?

WSRequestHandler.cpp
```diff
@@ -1343,6 +1345,141 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+
+        obs_data_set_double(data, "rotation", obs_sceneitem_get_rot(scene_item));
+
+        // investigate this and maybe convert so string in switch statment
+        // #define OBS_ALIGN_CENTER (0)
+        // #define OBS_ALIGN_LEFT   (1<<0)
+        // #define OBS_ALIGN_RIGHT  (1<<1)
+        // #define OBS_ALIGN_TOP    (1<<2)
+        // #define OBS_ALIGN_BOTTOM (1<<3)
+        obs_data_set_int(data, "alignment", obs_sceneitem_get_alignment(scene_item));
```
TStod: Alignment is an `int` that corresponds to the above `enum` definition. Should I break this out into a string based on cases or just offload this responsibility to next programmer?

WSRequestHandler.cpp
```diff
@@ -56,6 +56,8 @@ WSRequestHandler::WSRequestHandler(QWebSocket* client) :
     messageMap["SetSceneItemPosition"] = WSRequestHandler::HandleSetSceneItemPosition;
     messageMap["SetSceneItemTransform"] = WSRequestHandler::HandleSetSceneItemTransform;
     messageMap["SetSceneItemCrop"] = WSRequestHandler::HandleSetSceneItemCrop;
+    // This method name needs work. Perhaps just "GetSceneItemProperties"
```
haganbmj: Yeah, saying `Scene` twice seems redundant.

WSRequestHandler.cpp
```diff
@@ -1343,6 +1345,141 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+            vec2 bounds;
+            obs_sceneitem_get_bounds(scene_item, &bounds);
+            obs_data_set_float(bounds_data, "width", bounds.x);
+            obs_data_set_float(bounds_data, "height", bounds.y);
+            obs_data_set_obj(data, "bounds", bounds_data);
+        }
+
+        // add source width?
+        // add source height?
+        // add locked?
```
TStod: Source width? Source height? Locked?
Does it make sense to add these to this method?

WSRequestHandler.cpp
```diff
@@ -1343,6 +1345,141 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+
+        obs_sceneitem_release(scene_item);
+        req->SendOKResponse(data);
+    } else {
+        req->SendErrorResponse("specified scene item doesn't exist");
+    }
```
TStod: Should i put this else at the top in the if as a pseudo "early return"

WSRequestHandler.h
```diff
@@ -63,6 +63,7 @@ class WSRequestHandler : public QObject {
     static void HandleSetSceneItemPosition(WSRequestHandler* req);
     static void HandleSetSceneItemTransform(WSRequestHandler* req);
     static void HandleSetSceneItemCrop(WSRequestHandler* req);
+    // Add handler here
```
haganbmj: This is just a header file so all it really needs is the method name.
`static void HandleGetSceneItemSceneProperties(WSRequestHandler* req);` or whatever you settle on.

WSRequestHandler.cpp
TStod: Docs should probably say that this returns properties that are specific to a given scene and item (mostly not generic to an item)

WSRequestHandler.cpp
haganbmj: It's a bit rough around the edges, but something like this will work.
Just put them in whatever order best makes sense as that'll be what's used.

```
@return {Object} `something`
@return {String} `something.expand` description
@return {String} `something.dong` description
```

WSRequestHandler.cpp
haganbmj: Nested is probably cleaner. There's already another method using/returning nested properties so it's not like it's disrupting some unspoken rule.

WSRequestHandler.cpp
haganbmj: I kind of like the idea of putting it at the top for consistency. It would match the other returns for invalid params, scene, etc.

At some point I'd like to have a look at extracting those methods out since they're just getting rewritten over and over again, but that's something for another day.

WSRequestHandler.cpp
```diff
@@ -1343,6 +1344,156 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+ * @return {int} `position.y` The y position of the source from the top.
+ * @return {double} `rotation` The clockwise rotation of the item in degrees around the point of alignment.
+ * @return {int?string?} `alignment` The point on the source that the item is manipulated from.
+ * @return {double} `scale.x` The x-scale factor of the source.
+ * @return {double} `scale.y` The y-scale factor of the source.
+ * @return {int} `crop.top` The number of pixels cropped off the top of the source before scaling.
+ * @return {int} `crop.right` The number of pixels cropped off the right of the source before scaling.
+ * @return {int} `crop.bottom` The number of pixels cropped off the bottom of the source before scaling.
+ * @return {int} `crop.left` The number of pixels cropped off the left of the source before scaling.
+ * @return {bool} `visible` If the source is visible.
```
haganbmj: This whole project needs a non-passive rewrite for consistency.
I like using `visible` more than `render`, but I think they've been interchanged throughout.

Same out apply to `scale.x` and `cx`.

WSRequestHandler.cpp
haganbmj: I just wish I'd have better experience with c++ development before.
Visual Studio is a pain and I don't think I've ever had an easy time using makefiles. The only time I've actually been able to get builds + dll swapping going is at the office, but that's only because they have all the directories hardcoded and tell everyone to use the same installation directory structure.

TStod: I'm a bit out of practice but in principle the changes I'm working on are pretty easy to make.
Half the battle with programming is environment and setup IMO.

TStod:
@Palakis still looking for some feedback here. I'll have another PR ready shortly after for the setter

WSRequestHandler.cpp
```diff
@@ -1343,6 +1345,287 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+    obs_data_set_int(crop_data, "top", crop.top);
+    obs_data_set_int(crop_data, "right", crop.right);
+    obs_data_set_int(crop_data, "bottom", crop.bottom);
+    obs_data_set_obj(data, "crop", crop_data);
+
+    obs_data_set_bool(data, "visible", obs_sceneitem_visible(scene_item));
+
+    obs_data_t* bounds_data = obs_data_create();
+    obs_bounds_type bounds_type = obs_sceneitem_get_bounds_type(scene_item);
+    if (bounds_type == OBS_BOUNDS_NONE) {
```
TStod: Can you guys let me know what you think about the implementation of `bounds type`? The alternative would be to pass an integer instead of a string. Each has its pluses and minuses.
@haganbmj @Palakis

WSRequestHandler.cpp
```diff
@@ -1343,6 +1345,287 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+            new_bounds.y = obs_data_get_double(req_bounds, "y");
+        }
+        obs_sceneitem_set_bounds(scene_item, &new_bounds);
+        if (obs_data_has_user_value(req_bounds, "alignment")) {
+            const uint32_t bounds_alignment = obs_data_get_int(req_bounds, "alignment");
+            if (Utils::IsValidAlignment(bounds_alignment)) {
+                obs_sceneitem_set_bounds_alignment(scene_item, bounds_alignment);
+            }
+            // Send an error in the else statement?
+            // Append an error message to the response?
```
TStod: Thoughts on what to do when alignment has flags for `left` and `right` ? this is invalid so we should probably send an error message

Palakis:
Thanks @TStod for this patch, and thanks @haganbmj for the reviews. Sorry for the response delay.

WSRequestHandler.cpp
Palakis: Passing an integer in this case is more of a C/C++ trend, so I'd go for the string way here, which I think is more comfortable for JS and Python developers.

WSRequestHandler.cpp
Palakis: > This whole project needs a non-passive rewrite for consistency.

Can't agree more. I also believe that other aspects must be rethinked and rewritten, which can be the basis for work on 5.0.0.

WSRequestHandler.cpp
```diff
@@ -1343,6 +1345,286 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+ * @return {bool} `visible` If the source is visible.
+ * @return {String} `bounds.type` Type of bounding box.
+ * @return {int} `bounds.alignment` Alignment of the bounding box.
+ * @return {double} `bounds.x` Width of the bounding box.
+ * @return {double} `bounds.y` Height of the bounding box.
+ *
+ * @api requests
+ * @name GetSceneItemSceneProperties
+ * @category sources
+ * @since 4.1.3?
```
Palakis: `@since unreleased`

WSRequestHandler.cpp
haganbmj: Bounds type is an enum?
JS would just use whatever the json standard is, which in this case I think is the enum name `OBS_BOUNDS_NONE`

TStod: `obs_data_set_string(bounds_data, "type", "stretch");`
would become
`obs_data_set_string(bounds_data, "type", "OBS_BOUNDS_STRETCH");`
correct @haganbmj ?

WSRequestHandler.cpp
TStod: @haganbmj maybe throw this in an issue for 5.0.0 cleanup

TStod:
I think a weak map would be needed here if we wanted to simplify further

WSRequestHandler.cpp
Palakis: @TStod I'd agree with this

WSRequestHandler.cpp
Palakis: It should send an error, because it's confusing for users to fail silently if the request is malformed.

WSRequestHandler.cpp
```diff
@@ -1343,6 +1356,323 @@ void WSRequestHandler::HandleSetSceneItemCrop(WSRequestHandler* req) {
...
+            }
+            else if (new_bounds_type == "OBS_BOUNDS_SCALE_TO_HEIGHT") {
+                obs_sceneitem_set_bounds_type(scene_item, OBS_BOUNDS_SCALE_TO_HEIGHT);
+            }
+            else if (new_bounds_type == "OBS_BOUNDS_MAX_ONLY") {
+                obs_sceneitem_set_bounds_type(scene_item, OBS_BOUNDS_MAX_ONLY);
+            }
+            else {
+                bad_request = bad_bounds = true;
+                obs_data_set_string(bounds_error, "type", "invalid");
```
TStod: Building and returning an error object for invalid properties. @Palakis

TStod:
Tested this in debug mode and things seem to be working well!
@haganbmj @Palakis any idea how I can build it for production and install it manually in my OBS. The .dll seems to be only built for debug ATM?

Palakis:
Hi @TStod, I believe this is okay to merge. There isn't all the properties of a sceneitem, but your code is ready for future property additions.

@haganbmj I'd like to mark `SetSceneItemRender`, `SetSceneItemPosition`, `SetSceneItemCrop` and `SetSceneItemTransform` as deprecated requests. Do you think it is appropriate to do so?

Palakis:
@TStod Debug behaviour should be the same in Release. If you'd like to test your changes in a Release OBS Studio, you can copy the debug obs-websocket binary with debug dependencies (QtNetworkd.dll & QtWebSocketsd.dll) to your OBS installation.

TStod:
Sounds good! I'm with marking those for deprecated but perhaps waiting for this method to grow a little more before deprecating it is better.

@Palakis Thanks for the help! I'll give that a go. I assume I should also quarantine the current obs-websocket stuff. QtNetwork.dll & QtWebSockets.dll

haganbmj:
Yeah, deprecating seems reasonable. Can certainly add that to the docs.

TStod:
@Palakis Getting some errors in my legit OBS install but dont have time to debug ATM. Will try to figure it out later.
```
14:11:50.132: LoadLibrary failed for '../../obs-plugins/64bit/obs-websocket.dll': The specified module could not be found.
14:11:50.132:  (126)
14:11:50.133: Module '../../obs-plugins/64bit/obs-websocket.dll' not found
14:11:50.134: LoadLibrary failed for '../../obs-plugins/64bit/Qt5Networkd.dll': The specified module could not be found.
14:11:50.134:  (126)
14:11:50.134: Module '../../obs-plugins/64bit/Qt5Networkd.dll' not found
14:11:50.135: LoadLibrary failed for '../../obs-plugins/64bit/Qt5WebSocketsd.dll': The specified module could not be found.
14:11:50.135:  (126)
14:11:50.135: Module '../../obs-plugins/64bit/Qt5WebSocketsd.dll' not found
```

Issue #105, opened 2017-07-12
Broken macOS builds


ErikMinekus:
#### Issue type
Bug

#### Description
The recent changes to the Qt path seem to have broken the OS X build again. It looks like ```/Applications/OBS.app/Contents/Resources/bin/QtWebSockets``` does not correctly reference ```/Applications/OBS.app/Contents/Resources/bin/QtNetwork```, and instead loads ```/usr/local/Cellar/qt5/5.8.0_1/lib/QtNetwork.framework/Versions/5/QtNetwork```.
```
21:43:51.192: os_dlopen(/Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so->/Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so): dlopen(/Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so, 1): Library not loaded: /usr/local/Cellar/qt5/5.8.0_1/lib/QtNetwork.framework/Versions/5/QtNetwork
21:43:51.192:   Referenced from: /Applications/OBS.app/Contents/Resources/bin/QtWebSockets
21:43:51.192:   Reason: image not found
```

#### Steps to reproduce
Install https://obs-websocket-osx-builds.s3-eu-central-1.amazonaws.com/obs-websocket-add39cf-master.pkg and launch OBS Studio. The plugin fails to load.

#### Technical information
- **Operating System** : macOS Sierra 10.12.5
- **OBS Studio version** : 19.0.3

shawnbutts:
I'm looking forward to a osx release.

Palakis:
Tried to fix this in the latest builds. Try this: https://obs-websocket-osx-builds.s3-eu-central-1.amazonaws.com/obs-websocket-f995268-master.pkg

ErikMinekus:
It loads, but immediately crashes with the following stack trace:

```
0   libsystem_kernel.dylib        	0x00007fffd208ed42 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fffd217c457 pthread_kill + 90
2   libsystem_c.dylib             	0x00007fffd1ff4420 abort + 129
3   QtCore                        	0x0000000103efbb09 0x103ee5000 + 92937
4   QtCore                        	0x0000000103efd519 QMessageLogger::fatal(char const*, ...) const + 233
[...]

PR #141, opened 2017-10-19
macOS CI: Updated dependencies and namespaces

PicoCentauri:
Following the discussion of #105 the Qt version in the dependency installation is now hardcoded to version 5.9.2, which is the same as the main obs version for 20.1.0. Furthermore the names are changed from osx to macOS.


Palakis:
Thanks for this. I hope it will make better macOS builds of obs-websocket.

> the Qt version in the dependency installation is now hardcoded to version 1.9.2

Isn't it 5.9.2? 1.9.2 seems way too old.

PicoCentauri:
> Isn't it 5.9.2? 1.9.2 seems way too old.

Fixed

CI/install-dependencies-macos.sh
```diff
@@ -7,22 +7,26 @@ brew install ffmpeg
 brew install libav
 
 # qtwebsockets deps
-brew install qt5
+# qt latest
+#brew install qt5
+
+# qt 1.9.2
```
haganbmj: super minor nit: `5.9.2`

PicoCentauri: Argh my friend the typo... Should be fixed by now.

Palakis:
Thanks for this. Build seems to work fine.

PR #142, opened 2017-10-19
Fix saving replay buffer when no hotkey is set

RytoEX:
This commit makes saving the replay buffer not rely on whether or not a hotkey is set in OBS.

This is a temporary follow-up to PR #104, issue #90, and [OBS Mantis Bug 955](https://obsproject.com/mantis/view.php?id=955) until the OBS Frontend API gets a built-in function for this.  I am also working on that, which is why I called this temporary.  However, PRs to OBS core may take longer to get merged than a PR here, so I'm submitting this PR for now.

Also, I couldn't see an easy way to detect a success/fail from OBS on saving the replay, so I made the function here simply return an OK message.  The only way right now to detect if a reply was saved would be to manually check if the file exists.

I've compiled and tested this on Windows 10 Pro 64-bit.  I wasn't sure about the commit guidelines on this project, so please feel free to edit this as needed.

### DIFF
--- a/WSRequestHandler.cpp
+++ b/WSRequestHandler.cpp
@@ -780,13 +780,15 @@ void WSRequestHandler::HandleSaveReplayBuffer(WSRequestHandler* req) {
         return;
     }
 
-    obs_hotkey_t* hk = Utils::FindHotkeyByName("ReplayBuffer.Save");
-    if (hk) {
-        obs_hotkey_trigger_routed_callback(obs_hotkey_get_id(hk), true);
-        req->SendOKResponse();
-    } else {
-        req->SendErrorResponse("failed to save replay buffer");
-    }
+    calldata_t cd = {0};
+    obs_output_t* replay_output = obs_frontend_get_replay_buffer_output();
+    proc_handler_t* ph = obs_output_get_proc_handler(replay_output);
+    proc_handler_call(ph, "save", &cd);
+
+    req->SendOKResponse();
+
+    calldata_free(&cd);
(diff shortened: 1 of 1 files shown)


Palakis:
Thanks @RytoEX, triggering an RP save is better that way. Code is fine, as well as commit text.

Palakis:
On your end, do you believe it is okay to merge?

RytoEX:
My testing seemed to indicate that it functioned just fine.  I think it's okay to merge.  I'd keep an eye on [OBS Studio PR 1051](https://github.com/jp9000/obs-studio/pull/1051) (jp9000/obs-studio#1051) to see if/when that gets merged, but at least this fixes the remaining issue(s) with #90 and #104, specifically [this](https://github.com/Palakis/obs-websocket/pull/104#issuecomment-321981265):
> Handle the case where no hotkey is bound to save the replay buffer to disk: bind a dummy hotkey? return an error?

Palakis:
Even if the new technique doesn't rely on hotkeys, users will still have to set the hotkey when using RP through obs-websocket. Otherwise they'll get an error when trying to start the Replay Buffer.
Maybe there's a temporary way to trick OBS by adding a dummy hotkey, possibly invisible.

At least, your implementation is more future-proof.

Palakis:
The hotkey check counts how many items are in the bindings array of "save RP buffer" hotkey, whether these items are valid or not. I'll test if inserting an empty object in that array right before calling the RP start function does the trick without messing too much with OBS.

PR #149, opened 2017-10-27
Automatic function-scope OBS instance release + refactor variable names to camelCase

Palakis:
## Introduction
I've undergone a rewrite of some of obs-websocket's internals:
- Rewrite variable names to camelCase
- Use `obs_data_t*`, `obs_source_t*`, `obs_output_*` and others in special wrapper classes that automatically call their associated `release` function.

## Details
Current `release` code:
```cpp
void dummy() {
    obs_data_t* data_obj = obs_data_create();
    obs_data_set_string(data_obj, "key", "value);
    obs_data_release(data_obj);
}
```

New proposed `release` code:
```cpp
void dummy() {
    OBSDataAutoRelease dataObj = obs_data_create();
    obs_data_set_string(dataObj, "key", "value);
}
```

The goal of this rewrite is to unify code style, make easier the use of OBS' API and reduce the number of potential memory leaks. The auto-releases classes are named are based on `OBSRef` from libobs' C++ helpers in `obs.hpp`.

Note: `obs.hpp` already has classes that do automatic reference count management. However, because most libobs methods used in obs-websocket already call `addref` on the returned instance (hence increasing the reference count), using an OBSRef derivative results in a refcount offset up by one, potentially causing memory leaks.

Below is a comparison of current and new paradigms:

Current implementation:
```cpp
void dummy() {
    // obs_data_t* created. refcount = 1 
    obs_data_t* dataObj = obs_data_create();
    
    // obs_data_release decreases the refcount by one.
    // refcount is now 0 and instance can be destroyed.
    obs_data_release(dataObj);
}
```

Implemention using `obs.hpp`'s OBSRef derivatives 
```cpp
void dummy() {
    // obs_data_created. refcount = 1
    // OBSData captures the instance and calls addref() on it. refcount is now 2
    OBSData dataObj = obs_data_create();

    // End of function scope. dataObj is automatically destroyed.
    // OBSData's destructor calls release() on the instance, which decreases the refcount
    // by one. refcount is now 1, but the pointer is used only there and then the pointer will be lost
    // This causes a memory leak
}
```

Implemention using obs-websocket's OBS*AutoRelease derivatives of OBSRef 
```cpp
void dummy() {
    // obs_data_created. refcount = 1
    // OBSData captures the instance but doesn't call addref(). refcount is still 1.
    OBSDataAutoRelease dataObj = obs_data_create();

    // End of function scope. dataObj is automatically destroyed.
    // OBSData's destructor calls release() on the instance, which decreases the refcount
    // by one. refcount is now 0 and the instance is automatically destroyed. No memory leak.
}
```

This works fine with instances returned by most libobs methods. However, some don't addref on their result, and the only way to find out whether an OBSSource or OBSSourceAutoRelease should be used require digging through libobs's source code.
However, in most if not all cases, obs_data_t* instances should be used with OBSDataAutoRelease because most functions returning an obs_data_t* pointer call addref() before returning.

## Conclusion
I opened this as an internal PR because I'd like third-party reviews on some aspects:
- Is the new style easy enough to use and understand?
- Does it solves memory leaks in instances returned from libobs and easily avoids them when writing new code?

### DIFF
--- a/Config.cpp
+++ b/Config.cpp
@@ -32,6 +32,7 @@ with this program. If not, see <https://www.gnu.org/licenses/>
 #define PARAM_SALT "AuthSalt"
 
 #include "Config.h"
+#include "Utils.h"
 
 Config* Config::_instance = new Config();
 
@@ -46,24 +47,24 @@ Config::Config() :
     SettingsLoaded(false)
 {
     // OBS Config defaults
-    config_t* obs_config = obs_frontend_get_global_config();
-    if (obs_config) {
-        config_set_default_bool(obs_config,
+    config_t* obsConfig = obs_frontend_get_global_config();
+    if (obsConfig) {
+        config_set_default_bool(obsConfig,
             SECTION_NAME, PARAM_ENABLE, ServerEnabled);
-        config_set_default_uint(obs_config,
+        config_set_default_uint(obsConfig,
             SECTION_NAME, PARAM_PORT, ServerPort);
 
-        config_set_default_bool(obs_config,
+        config_set_default_bool(obsConfig,
             SECTION_NAME, PARAM_DEBUG, DebugEnabled);
-        config_set_default_bool(obs_config,
+        config_set_default_bool(obsConfig,
             SECTION_NAME, PARAM_ALERT, AlertsEnabled);
 
-        config_set_default_bool(obs_config,
+        config_set_default_bool(obsConfig,
             SECTION_NAME, PARAM_AUTHREQUIRED, AuthRequired);
-        config_set_default_string(obs_config,
-            SECTION_NAME, PARAM_SECRET, Secret);
-        config_set_default_string(obs_config,
-            SECTION_NAME, PARAM_SALT, Salt);
+        config_set_default_string(obsConfig,
+            SECTION_NAME, PARAM_SECRET, qstring_data_copy(Secret));
+        config_set_default_string(obsConfig,
+            SECTION_NAME, PARAM_SALT, qstring_data_copy(Salt));
     }
 
     mbedtls_entropy_init(&entropy);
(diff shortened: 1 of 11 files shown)


haganbmj:
I was referring to the json request/response objects when I mentioned camelCase in case that wasn't clear.
I know snake case isn't uncommon in cpp for naming things if that's some style you like.

Palakis:
> I was referring to the json request/response objects when I mentioned camelCase in case that wasn't clear.

I'm completely aware of this. This PR has nothing to do with 5.0.0.

haganbmj:
Cool, just wanted to make sure.

Palakis:
Results so far: no memory leaks on this branch as far as my manual testing went. Gonna do more testing to make sure of that.

Palakis:
One thing bothers me currently: `OBSDataAutoRelease` is quite a long name for a class that's used extensively in the source code. Maybe `OBSDataRef` is a more suitable shorter name.

Palakis:
> One thing bothers me currently: OBSDataAutoRelease is quite a long name for a class that's used extensively in the source code. Maybe OBSDataRef is a more suitable shorter name.

Same bothers for the other *DataRelease classes, too.

TStod:
Solving memory leaks where ever possible is obviously ideal. In general the pattern we use for simple library additions should be leak free. I like the changes you proposed.

Long names don't bother me quite so much because of tab completion.

haganbmj:
80char-line-length

Long variable names make it harder to adhere to 80 char line width! How will I ever see full lines on my screen! /sarcasm

I personally don't mind long names because it makes it easier to read and to come back to later. Line length should be more of a guideline in terms of code complexity to prevent excessive nesting.

Palakis:
Thanks for your feedback. Gonna merge this after a last batch of manual (sigh) tests.

TStod:
@Palakis I am noticing a bug with `Utils::GetSceneItemFromName` where if you pass a valid scene and a valid scene item that is not in that scene, the scene item is still returned. Expected behavior is for it to not be returned because the scene item is not present within the specified scene.

PR #158, opened 2017-11-14
Request: Create folder for 'SetRecordingFolder' if it does not exist

DungFu:
When using 'SetRecordingFolder', it will fail if the folder does not already exist. If one wants to save recordings to changing folders (such as a different folder for each game), this is not possible without running an extra program/script.

This pull request changes obs-websocket to try to create the directory (and all intermediaries) when using 'SetRecordingFolder' to solve this issue.

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -367,9 +367,8 @@ const char* Utils::GetRecordingFolder() {
 }
 
 bool Utils::SetRecordingFolder(const char* path) {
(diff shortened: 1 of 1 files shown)


Palakis:
Thanks for this. Why did you revert your changes?

DungFu:
whoops, I wanted to make another pull request and reverted on master :P

I'll make a new pull request, my bad.

PR #160, opened 2017-11-14
Added the local_file path to the Browsersources

PicoCentauri:
OBS distinguishes differentiates between the url and the local_file. This is now taken into account.


Palakis:
Looks good to me. Thanks for this!

PR #162, opened 2017-11-14
Add support for filename formatting get/set

DungFu:
Adds the ability to get/set the filename formatting string parameter. Tested get/set and both work correctly (there is no validation on setting besides checking if the string is not empty).

Feedback is appreciated :)


Palakis:
@DungFu Thanks for this. Gonna have a look at it in a few hours.

src/WSRequestHandler_General.cpp
```diff
@@ -124,3 +124,42 @@ void WSRequestHandler::HandleAuthenticate(WSRequestHandler* req) {
...
+
+/**
+ * Set the filename formatting string
+ *
+ * @param {String} `filename-formatting` Filename formatting string to set.
+ *
+ * @api requests
+ * @name SetFilenameFormatting
+ * @category general
+ */
```
Palakis: don't forget `@since unreleased` in request types you introduce.

PR #163, opened 2017-11-14
Create folder for 'SetRecordingFolder' if it does not exist

DungFu:
When using 'SetRecordingFolder', it will fail if the folder does not already exist. If one wants to save recordings to changing folders (such as a different folder for each game), this is not possible without running an extra program/script.

This pull request changes obs-websocket to try to create the directory (and all intermediaries) when using 'SetRecordingFolder' to solve this issue.

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -367,9 +367,8 @@ const char* Utils::GetRecordingFolder() {
 }
 
 bool Utils::SetRecordingFolder(const char* path) {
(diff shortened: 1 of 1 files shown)


DungFu:
Any chance someone could take a look at this? It would be really nice to have this merged in for an app I am making :)

Palakis:
@DungFu Looks good to me. Thanks for this!

PR #191, opened 2018-02-23
fixing get item from item bug and duplicate item bug

TStod:


TStod:
@Palakis these should go in ASAP now that the id refactor is already merged

PR #196, opened 2018-03-09
Workaround/fix for Control-Z in qstring_data_copy

dx9s:
Changed const char * to char *

Perhaps not the BEST solution!? I am open to better!?

--Doug (dx9s)

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -28,11 +28,12 @@ with this program. If not, see <https://www.gnu.org/licenses/>
 
 Q_DECLARE_METATYPE(OBSScene);
 
-const char* qstring_data_copy(QString value) {
+char* qstring_data_copy(QString value) {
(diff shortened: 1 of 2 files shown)


Palakis:
@dx9s Thanks for this. I'll have a look.

Palakis:
@dx9s I'm about to disappoint you but I fixed it in `master` (https://github.com/Palakis/obs-websocket/commit/e168b78cc13cb0a44d4bb32c795eae2b3e884eb9).
`config_set_default_string` and `config_set_string` call a variant of `strdup` internally and handle their strings properly, contrary to our flawed `qstring_data_copy` which did unnecessary and buggy string copying.

dx9s:
Sounds fine.. I was sure there should be a cleaner way to do it.. I am not verse in all thing Qt .. nor that current on C++ . BUT if it's fixed.. cool I'll poke at the solution and learn.

PR #205, opened 2018-03-26
CI and CMake improvements

RytoEX:
This PR is a series of general CI and CMake improvements.

This PR adds support for the build config/type "RelWithDebInfo", which is what OBS uses for its official releases, and it sets the default build config to that.  This PR has also shown a reduction in CI build times of 1-5 minutes depending on CI service, CI status/availability, and whether a full OBS build is needed.

I still have some macOS specific CI/CMake changes I'd like to work on, but I figured I'd submit these first, so this PR can be considered separately from those later changes.  These changes build locally and on CI, and obs-websocket seems to work when run through some simple tests with the t2t2 web client.


RytoEX:
I think the AppVeyor failure here is an issue with the caching.  It didn't force a rebuild of OBS, but it needed to because the build config changed.  I could probably try to cache the last build config or check if the current build config directory exists to try to avoid that issue.

PR #207, opened 2018-04-13
macOS CI updates

RytoEX:
This PR does several things:
* Update Packages to 1.2.3 to keep CI running correctly
* Update Qt to latest in brew (this is what OBS Studio CI does)
* Build the most recent OBS Studio git tag (as of this date 21.1.1)
* Move the obs-studio build task to a separate script (creates a separate section in CI logs, and lets user skip the obs-studio build script if building locally with scripts)

I have more macOS CI changes in the pipeline, but I'm struggling with a few of them, so I thought I'd split these off since they seem stable.

The obs-studio build script could probably use a bit of directory/path checking to make sure it's changing into the correct directory, but that should only be a concern if using the scripts locally, and I don't have access to a Mac to test locally.

On the note of keeping up with Packages updates, I did notice that obs-studio does something different with that [here](https://github.com/obsproject/obs-studio/blob/3040ff6347d006e52c90b5be365cae7a04aa07d6/CI/install-dependencies-osx.sh#L12).  Is this something that could be done here, or is Packages handled differently here for a reason?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,9 @@ matrix:
   - os: osx
     env: _macos_build
     osx_image: xcode8.3
-    before_install: "./CI/install-dependencies-macos.sh"
+    before_install:
+    - "./CI/install-dependencies-macos.sh"
+    - "./CI/install-build-obs-macos.sh"
     script: "./CI/build-macos.sh"
     after_success: 
     - ./CI/package-macos.sh
--- a/CI/install-build-obs-macos.sh
+++ b/CI/install-build-obs-macos.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+set -ex
+
+# Build obs-studio
+cd ..
+git clone https://github.com/obsproject/obs-studio
+cd obs-studio
+OBSLatestTag=$(git describe --tags --abbrev=0)
+git checkout $OBSLatestTag
+mkdir build && cd build
+cmake .. \
+  -DDISABLE_PLUGINS=true \
+  -DCMAKE_PREFIX_PATH=/usr/local/opt/qt/lib/cmake \
+&& make -j4
--- a/CI/install-dependencies-macos.sh
+++ b/CI/install-dependencies-macos.sh
@@ -8,10 +8,7 @@ brew install libav
 
 # qtwebsockets deps
 # qt latest
-#brew install qt5
-
-# qt 5.9.2
-brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/2b121c9a96e58a5da14228630cb71d5bead7137e/Formula/qt.rb
+brew install qt5
 
 #echo "Qt path: $(find /usr/local/Cellar/qt5 -d 1 | tail -n 1)"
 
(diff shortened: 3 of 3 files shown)


RytoEX:
I updated this so that it always builds the latest obs-studio git tag.  I still haven't gotten build caching working like it does on AppVeyor, and I'm not sure how to solve that problem.  I did start a branch that implements ccache instead, and that does cut build times down by a minute or more, but it doesn't enable completely skipping the obs-studio build.  I can either add the ccache commit to this PR, or file it as a separate PR later.  Thoughts?

RytoEX:
I actually have a local branch with updates to `BUILDING.md` already made.  I just wasn't sure if @Palakis wanted them rolled into here somewhere or submitted separately.

About directory checking, I'm not really concerned about `install-dependencies-macos.sh` so much as I am about `install-build-obs-macos.sh` just making dirs and trying to run cmake in the wrong place.  I guess user beware?

About Packages, I asked that exact question in [my original comment above](https://github.com/Palakis/obs-websocket/pull/207#issue-181591277).  I'm not familiar enough with macOS or Packages to understand the difference between how the obs-studio CI installs Packages versus how it's done here.  The code changes are easy enough to make though.  I could probably make a quick PR to mimic what obs-studio's CI does and ask Palakis to merge that into a test OSX branch to test the deploy steps to avoid cluttering up any of the meaningful branches.

wherget:
My usual approach is that PRs should be complete in the sense that everything should be stable again after the merge. As it stands, the `BUILDING.md` changes left out "break" manually building after this has landed. I would therefore have included the changes.

I was also referring to `install-build-obs-macos.sh`, sorry if that didn't come across. In my runs that once properly failed for me due to `set -x`. User beware seems enough.

Having now had a look at how obs-studio are handling the Packages dependency (to wit, keeping their own version of the `Packages.pkg` on their AWS mirror) I think they're also going for the "stable environment" approach. While it's not for me to say, I would either leave it as is (expect `1.2.3` in the upstream DMG) or download OBS's mirror (for consistency reasons). Both solutions are implicitly placing trust on third parties to have the right things in the right place.

Thanks again for your work!

wherget:
OK, while obs-websocket _built_ with your changes, it didn't actually _work_ (as I found out yesterday) due to library shenanigans (as I found out today). There turn out to be a surprising number of places where the library paths are baked in and proliferate through `dyld`. But I got it working. Now the question is whether we should roll my changes ( into here, since they fix the CI __result__ (whereas the current state only fixes the CI __process__)? Or should I open a separate PR once this one has landed? Rolling the changes into here probably means you @RytoEX would be pulling in my changes into your repo, if you're OK with that. Would you like a PR?

Considering the `Packages.pkg` thing, I just found [obs-ndi](https://github.com/Palakis/obs-ndi/blob/master/CI/install-dependencies-osx.sh) also grabs the obs-nightly mirror. I would presume @Palakis to be fine with that solution then.

RytoEX:
@wherget 
If you already have a solution for fixing library errors introduced by my work here, feel free to PR against my branch and I'll review it there.  That would save me the time and trouble of hunting it down, and it would prevent a broken build from being produced once (if) merged.

Interesting about the difference in setup between this and obs-ndi.  I'd still like @Palakis to chime into confirm that that's okay here as well.  I may be unavailable to make that change for a few days anyway.

Palakis:
Yeah I don't mind doing the same as in obs-ndi and use the Packages.dmg provided on obs-nightly. @RytoEX you can add that fix I believe.

PR #211, opened 2018-04-29
MacOS Build fixes

wherget:
This PR adds some fixes mostly meant for people that build obs-websocket themselves using the CI scripts, which boil down to "don't fail when running the script a second time because the first try failed".

Included are some minor cleanups for `dyld` runtime path handling, making sure no build environment path artifacts are baked into the binaries produced. The current behavour may screw library loading up for people that happen to have something installed in the same paths as the travis environment, which basically boils down to having Qt 5.10.1 installed via homebrew (whether or not this is due to building obs-websocket yourself).

I tried to make the commit messages as explanatory as possble. If you have any further comments please let me know.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -22,6 +22,8 @@ sudo make install
 ```
 
 ## OS X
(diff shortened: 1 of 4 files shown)


JulienBreux:
Thanks!

PR #221, opened 2018-05-27
Request types for filter management

Palakis:
- GetSourceFilters
- AddFilterToSource
- RemoveFilterFromSource
- SetSourceFilterSettings
- ReorderSourceFilter
- MoveSourceFilter

### DIFF
--- a/src/WSRequestHandler_Sources.cpp
+++ b/src/WSRequestHandler_Sources.cpp
@@ -1313,3 +1313,314 @@ void WSRequestHandler::HandleGetSpecialSources(WSRequestHandler* req) {
 
     req->SendOKResponse(response);
 }
+
+/**
+* List filters applied to a source
+*
+* @param {String} `sourceName` Source name
+*
+* @return {Array of Objects} `filters` List of filters for the specified source
+* @return {String} `filters.*.type` Filter type
+* @return {String} `filters.*.name` Filter name
+* @return {Object} `filters.*.settings` Filter settings
+*
+* @api requests
+* @name GetSourceFilters
+* @category sources
+* @since unreleased
+*/
+void WSRequestHandler::HandleGetSourceFilters(WSRequestHandler* req)
+{
+	if (!req->hasField("sourceName")) {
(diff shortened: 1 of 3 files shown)


Palakis:
@haganbmj @TStod @RytoEX May I ask one of you to have a look at this, please?

src/WSRequestHandler_Sources.cpp
```diff
@@ -1313,3 +1313,314 @@ void WSRequestHandler::HandleGetSpecialSources(WSRequestHandler* req) {
...
+		req->SendErrorResponse("filter name already taken");
+		return;
+	}
+
+	OBSSourceAutoRelease filter = obs_source_create_private(filterType, filterName, filterSettings);
+	if (!filter) {
+		req->SendErrorResponse("filter creation failed");
+		return;
+	}
+	if (obs_source_get_type(filter) != OBS_SOURCE_TYPE_FILTER) {
```
haganbmj: Out of curiosity, what other types could it be at this point?

TStod:
IIRC OBS was investigating filters on a per scene item basis (instead of a per source basis) I'm unsure of the status of that change and will do some digging as it will inform our decisions.

src/WSRequestHandler_Sources.cpp
Palakis: @haganbmj Filters are sources, which type is one of the `OBS_SOURCE_TYPE_*` constants.

Palakis: @haganbmj 
> what other types could it be at this point?

Actually I don't know. I just want to add another security to make sure the source created is indeed a filter.

haganbmj: Yeah, I saw elsewhere in this commit where filters are `OBSSourceAutoRelease`, so makes sense to take precaution.

RytoEX:
Should I look at anything in particular?

Llorx:
How is this going right now?

vicegold:
Any update on filter control via obs-websocket?

PR #240, opened 2018-09-25
Can add and remove video sources

kraslin:
Websocket clients can now add a completely new video source to a scene, also remove a video source from a scene.
Pull request would be against master but master doesn't seem to work.


kraslin:
Also fixed string equality check while deducing bounds type in SetSceneItemProperties

PR #241, opened 2018-09-27
Add wget to install-dependencies-macos.sh

caseymrm:
wget is not installed by default, so when doing the build on a fresh machine, brew install it first


Palakis:
Thanks for this!

PR #245, opened 2018-10-17
CI: Fix QT 5.10 not building under macOS 10.13+

PatTheMav:
As described by in https://github.com/Homebrew/homebrew-core/issues/27095, it is impossible to build QT 5.10 on any macOS newer than macOS 10.12 due to it using C++ 17 by default.

As explained in the issue, the changes outlined in https://github.com/Homebrew/homebrew-core/pull/27139 need to be applied to the Homebrew formula. With those QT 5.10 builds successfully and the result obs-websocket build at least loads without issue in OBS 22.0.3 under macOS 10.14.

### DIFF
--- a/CI/macos/qt.rb
+++ b/CI/macos/qt.rb
@@ -18,7 +18,7 @@ class Qt < Formula
 
   option "with-docs", "Build documentation"
   option "with-examples", "Build examples"
-  
+
   deprecated_option "with-mysql" => "with-mysql-client"
 
   # OS X 10.7 Lion is still supported in Qt 5.5, but is no longer a reference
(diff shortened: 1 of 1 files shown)


Palakis:
Thanks for this!

PatTheMav:
@Palakis It's the least I could do! However even though the library is loaded without issue, `WSServer::onNewConnection` is never triggered even though a connection attempt is made. I added additional `blog` statements in a local build and even though I tried to connect locally via `cURL` there was no output in OBS' log.

If a close OBS, the client reports that the connection is closed by the remote host, so a connection is seemingly established, however to data flows back or forth. My C++ knowledge is a bit rusty, so I'm currently at a loss what the exact issue might be.

Issue #238, opened 2018-09-21
obs-websocket does not load on mac


naychrist:
##### Issue type
Bug report

##### Description
OBS launches but the plugin does not load (nothing appears under tools). launching from terminal or viewing log I can see this is due to a qt version mismatch:

```
error: os_dlopen(/Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so->/Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so): dlopen(/Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so, 1): Library not loaded: @rpath/QtWidgets
  Referenced from: /Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so
  Reason: Incompatible library version: obs-websocket.so requires version 5.11.0 or later, but QtWidgets provides version 5.10.0
```

After trying with the downloaded binary I installed qt 5.11 using brew and built obs-websocket from scratch as per instructions. This compiles fine but after install I have the same issue, even though I can verify that qtwebsockets is version 5.11.2 at:
`/usr/local/opt/qt/lib/QtWebSockets.framework`

##### Steps to reproduce and other useful info
Launch latest OBS after installing latest obs-websocket on OSX, from downloaded
[2018-09-21 17-52-24.txt](https://github.com/Palakis/obs-websocket/files/2404429/2018-09-21.17-52-24.txt)
 or built installer.

##### Technical information
- **Operating System** : Mac OSX High Sierra 10.13.6
- **OBS Studio version** : 22.0.3

##### Development Environment
I have XCode 9.4.1 installed but am using the default build scripts in the repo.

naychrist:
Just a note that when building both OBS and obs-websockets with qt 5.11.2 I actually get a different error than a qt mismatch when launching OBS:

```
error: os_dlopen(/Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so->/Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so): dlopen(/Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so, 1): Library not loaded: @rpath/QtWebSockets
  Referenced from: /Library/Application Support/obs-studio/plugins/obs-websocket/bin/obs-websocket.so
  Reason: image not found
```

However, it is installed before building and the path seems to be set correctly set in the build_macos.sh script:
[...]

PR #246, opened 2018-10-24
4.4 macOS build system update

PatTheMav:
As discussed in https://github.com/Palakis/obs-websocket/issues/238 this PR updates the build scripts for macOS.

I added some dependency checks and cleaned-up Homebrew handling. The build-files also use the same QT build as obs-studio itself (checked their CI scripts, they install from the same remote formula). This speeds up building (as QT is "poured from a bottle") and ensures compatibility - for now.

I've also updated the bundled QT5 formula to match the remote one, but also added the necessary patches to make it build on macOS 10.13+.

Additionally the scripts now "pin" qt to the installed version, to avoid `brew upgrade` from upgrading it to 5.12.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,12 +34,12 @@ matrix:
 
   - os: osx
     env: _macos_build
-    osx_image: xcode8.3
+    osx_image: xcode9.4
     before_install:
     - "./CI/install-dependencies-macos.sh"
     - "./CI/install-build-obs-macos.sh"
     script: "./CI/build-macos.sh"
-    after_success: 
+    after_success:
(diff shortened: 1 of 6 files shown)


PatTheMav:
So sorry for the multiple commits, but I'm learning here as I go and I _believe_ I managed to fix most CI issues. Alas obs-websocket itself will fail to build due to the issue I mentioned in https://github.com/Palakis/obs-websocket/commit/97109087a47b0bc9db4114fc099186ec13db031c.

Palakis:
@PatTheMav thanks!

PR #272, opened 2019-01-20
Support item_transform signal

BarryCarlyon:
##### Issue type
Feature request

##### Description
Current when a source on a scene is resized or moved there is no event to notify socket subscribers that this occured. OBS does emit `item_transform` for this.

Clone OnSceneItemAdd to a new function to support the `item_transform` signal. Emit as `SourceItemChanged`

Basically here I've just duplicated and adjusted `OnSceneItemAdd` to match the existing format.

Personally I'd extend to emit the details about the scene. But this doesn't match what the other events do in OBS-Websocket in this group.

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -553,3 +553,77 @@ bool Utils::SetFilenameFormatting(const char* filenameFormatting) {
 	config_save(profile);
 	return true;
 }
+
+obs_data_t* Utils::GetSceneItemPropertiesData(obs_sceneitem_t* sceneItem) {
+	OBSDataAutoRelease data = obs_data_create();
+
+	OBSDataAutoRelease posData = obs_data_create();
+	vec2 pos;
+	obs_sceneitem_get_pos(sceneItem, &pos);
+	obs_data_set_double(posData, "x", pos.x);
+	obs_data_set_double(posData, "y", pos.y);
+	obs_data_set_int(posData, "alignment", obs_sceneitem_get_alignment(sceneItem));
+	obs_data_set_obj(data, "position", posData);
+
+	obs_data_set_double(data, "rotation", obs_sceneitem_get_rot(sceneItem));
+
+	OBSDataAutoRelease scaleData = obs_data_create();
+	vec2 scale;
+	obs_sceneitem_get_scale(sceneItem, &scale);
+	obs_data_set_double(scaleData, "x", scale.x);
+	obs_data_set_double(scaleData, "y", scale.y);
+	obs_data_set_obj(data, "scale", scaleData);
+
+	OBSDataAutoRelease cropData = obs_data_create();
+	obs_sceneitem_crop crop;
+	obs_sceneitem_get_crop(sceneItem, &crop);
+	obs_data_set_int(cropData, "left", crop.left);
+	obs_data_set_int(cropData, "top", crop.top);
+	obs_data_set_int(cropData, "right", crop.right);
+	obs_data_set_int(cropData, "bottom", crop.bottom);
+	obs_data_set_obj(data, "crop", cropData);
+
+	obs_data_set_bool(data, "visible", obs_sceneitem_visible(sceneItem));
+
+	OBSDataAutoRelease boundsData = obs_data_create();
+	obs_bounds_type boundsType = obs_sceneitem_get_bounds_type(sceneItem);
+	if (boundsType == OBS_BOUNDS_NONE) {
+		obs_data_set_string(boundsData, "type", "OBS_BOUNDS_NONE");
+	}
+	else {
+		switch (boundsType) {
+		case OBS_BOUNDS_STRETCH: {
+			obs_data_set_string(boundsData, "type", "OBS_BOUNDS_STRETCH");
+			break;
+		}
+		case OBS_BOUNDS_SCALE_INNER: {
+			obs_data_set_string(boundsData, "type", "OBS_BOUNDS_SCALE_INNER");
+			break;
+		}
+		case OBS_BOUNDS_SCALE_OUTER: {
+			obs_data_set_string(boundsData, "type", "OBS_BOUNDS_SCALE_OUTER");
+			break;
+		}
+		case OBS_BOUNDS_SCALE_TO_WIDTH: {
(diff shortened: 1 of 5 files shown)


src/WSEvents.cpp
```diff
@@ -911,6 +915,38 @@ void WSEvents::OnSceneItemVisibilityChanged(void* param, calldata_t* data) {
...
+	const char* sceneName =
+		obs_source_get_name(obs_scene_get_source(scene));
+	const char* sceneItemName =
+		obs_source_get_name(obs_sceneitem_get_source(sceneItem));
+
+	OBSDataAutoRelease fields = obs_data_create();
+	obs_data_set_string(fields, "scene-name", sceneName);
+	obs_data_set_string(fields, "item-name", sceneItemName);
+
+	instance->broadcastUpdate("SourceItemChanged", fields);
```
Palakis:
```suggestion
	instance->broadcastUpdate("SceneItemTransformChanged", fields);
```

src/WSEvents.cpp
```diff
@@ -911,6 +915,38 @@ void WSEvents::OnSceneItemVisibilityChanged(void* param, calldata_t* data) {
...
 }
 
+/**
+ * An item's transfrom has been changed.
+ *
+ * @return {String} `scene-name` Name of the scene.
+ * @return {String} `item-name` Name of the item in the scene.
+ *
+ * @api events
+ * @name SourceItemChanged
```
Palakis: It would be best to keep scene item event names in the same fashion as others.
```suggestion
 * @name SceneItemTransformChanged
```

src/WSEvents.cpp
```diff
@@ -911,6 +915,38 @@ void WSEvents::OnSceneItemVisibilityChanged(void* param, calldata_t* data) {
...
+
+	const char* sceneName =
+		obs_source_get_name(obs_scene_get_source(scene));
+	const char* sceneItemName =
+		obs_source_get_name(obs_sceneitem_get_source(sceneItem));
+
+	OBSDataAutoRelease fields = obs_data_create();
+	obs_data_set_string(fields, "scene-name", sceneName);
+	obs_data_set_string(fields, "item-name", sceneItemName);
+
```
Palakis: Please add the transform state in the event properties. You can reuse the schema used in the `GetSceneItemProperties` RPC method.

src/WSEvents.cpp
```diff
@@ -911,6 +915,38 @@ void WSEvents::OnSceneItemVisibilityChanged(void* param, calldata_t* data) {
...
+/**
+ * An item's transfrom has been changed.
+ *
+ * @return {String} `scene-name` Name of the scene.
+ * @return {String} `item-name` Name of the item in the scene.
+ *
+ * @api events
+ * @name SourceItemChanged
+ * @category sources
+ * @since 4.5.1
```
Palakis: `@since` will be updated on new releases.

```suggestion
 * @since unreleased
```

src/WSEvents.cpp
BarryCarlyon: That is of course unless it doesn't trigger when you transform a non active scene...

Palakis: @BarryCarlyon It's fine, still good to have.

src/WSEvents.cpp
```diff
@@ -911,6 +915,106 @@ void WSEvents::OnSceneItemVisibilityChanged(void* param, calldata_t* data) {
...
+			break;
+		}
+		}
+		obs_data_set_int(boundsData, "alignment", obs_sceneitem_get_bounds_alignment(sceneItem));
+		vec2 bounds;
+		obs_sceneitem_get_bounds(sceneItem, &bounds);
+		obs_data_set_double(boundsData, "x", bounds.x);
+		obs_data_set_double(boundsData, "y", bounds.y);
+	}
+	obs_data_set_obj(fields, "bounds", boundsData);
```
Palakis: To avoid redundant code, it would be best to put this chunk of code into its own `Utils` function so that both this method and `GetSceneItemProperties` use it.

BarryCarlyon: I think this should do it. My C is very rusty. And I don't have this machine setup right for proper OBS testing just this second. (Since this looked like a straight forward copy/paste fudge job/patch).

Palakis:
@BarryCarlyon Could you please rebase your branch onto the latest `4.x-current` branch please? CI has been fixed on macOS. Once you do that I will review this.

Palakis:
@BarryCarlyon Alright nevermind, your PR is eligible for edits by maintainers so I rebased it.

src/Utils.cpp
```diff
@@ -553,3 +553,77 @@ bool Utils::SetFilenameFormatting(const char* filenameFormatting) {
 	config_save(profile);
 	return true;
 }
+
+obs_data_t* Utils::GetSceneItemPropertiesData(obs_sceneitem_t* sceneItem) {
+	OBSDataAutoRelease data = obs_data_create();
```
Palakis: It's a bit risky to do this. When leaving the current scope, `data` will be automatically released and destroyed because its reference count was `1`. `OBSDataAutoRelease` only performs releases and doesn't add new references.

```suggestion
	obs_data_t* data = obs_data_create();
```

src/Utils.cpp
```diff
@@ -553,3 +553,77 @@ bool Utils::SetFilenameFormatting(const char* filenameFormatting) {
...
+	obs_data_set_bool(data, "visible", obs_sceneitem_visible(sceneItem));
+
+	OBSDataAutoRelease boundsData = obs_data_create();
+	obs_bounds_type boundsType = obs_sceneitem_get_bounds_type(sceneItem);
+	if (boundsType == OBS_BOUNDS_NONE) {
+		obs_data_set_string(boundsData, "type", "OBS_BOUNDS_NONE");
+	}
+	else {
+		switch (boundsType) {
+		case OBS_BOUNDS_STRETCH: {
```
Palakis: Please indent cases within switch blocks as well

Palakis:
@BarryCarlyon Do you want to carry out these fixes or should I do it if you're busy?

BarryCarlyon:
@Palakis sounds like a good idea. I am a little busy currently and I'm not native to C++ hence how stumbly this has been

PR #279, opened 2019-02-13
Add caption handler

derrod:
Added a handler that allows sending text as captions through OBS' built-in caption functionality.

I wasn't sure where to add it, so I just put it into streaming.

Also I hacked together a little demo site: https://rodney.io/stuff/obscaptions/ Just click the mic and it will (try to) connect to OBS and start sending captions. It would need a lot more work to be useful, but it works.


derrod:
I fixed CI for windows and (hopefully) macOS. Linux is using OBS from the PPA which appears to be built without captions support for some reason.

PR #295, opened 2019-04-19
Fix call name GetSourcesTypesList -> GetSourceTypesList

Rosuav:
The API call listed in the documentation doesn't work (tested on 4.3.1),
but without the 's', it works fine. Update documentation to match code.


Palakis:
@Rosuav Thanks for this!

PR #315, opened 2019-05-02
chore: update test dependencies

Lange:
`obs-websocket-js` now includes typedefs directly in the package, so `obs-websocket-js-types` is deprecated and no longer necessary.

`obs-websocket-js` requires a pretty modern version of TypeScript, so I've also updated the `typescript` and `ts-node` dependencies to match.


Lange:
The Azure Pipelines build failure appears to be unrelated.

Issue #296, opened 2019-04-19
Way to query OBS canvas size?


Rosuav:
##### Issue type
Question, possible feature request

##### Description
Is there a way to find out the OBS canvas size? Placement of all widgets is by pixels but the canvas size doesn't seem to be given anywhere.

##### Technical information
- **Operating System** : Linux
- **OBS Studio version** : 19.0.3, 23.0.2

Palakis:
@Rosuav it is indeed a feature request. This will potentially be released in 4.7, unless someone can take care of this for inclusion in 4.6.

Rosuav:
Okay, cool. At least I don't have egg on my face for being blind (which is a weird mixed metaphor, but where's the fun in unmixing them?). For now, I'll just hard-code it in my app.

Rosuav:
Closed by the above PR

PR #319, opened 2019-05-08
Create an endpoint to view basic video info eg canvas size

Rosuav:
I'm not sure that this is the best way to do it, nor if it's the best place to put this code. But it's a start.

Implements #296.

### DIFF
--- a/src/WSRequestHandler_General.cpp
+++ b/src/WSRequestHandler_General.cpp
@@ -185,10 +185,10 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
 /**
  * Get basic OBS video information
  * 
- * @return {Number} `baseWidth` Base (canvas) width
- * @return {Number} `baseHeight` Base (canvas) height
- * @return {Number} `outputWidth` Output width
- * @return {Number} `outputHeight` Output height
+ * @return {int} `baseWidth` Base (canvas) width
+ * @return {int} `baseHeight` Base (canvas) height
+ * @return {int} `outputWidth` Output width
+ * @return {int} `outputHeight` Output height
  * @return {String} `scaleType` Scaling method used if output size differs from base size
  * @return {double} `fps` Frames rendered per second
  * @return {String} `videoFormat` Video color format
--- a/src/WSRequestHandler_General.cpp
+++ b/src/WSRequestHandler_General.cpp
@@ -189,7 +189,11 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
  * @return {Number} `baseHeight` Base (canvas) height
  * @return {Number} `outputWidth` Output width
  * @return {Number} `outputHeight` Output height
+ * @return {String} `scaleType` Scaling method used if output size differs from base size
  * @return {double} `fps` Frames rendered per second
+ * @return {String} `videoFormat` Video color format
+ * @return {String} `colorSpace` Color space for YUV
+ * @return {String} `colorRange` Color range (full or partial)
  * 
  * @api requests
  * @name GetVideoInfo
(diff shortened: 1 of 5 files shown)


src/WSRequestHandler_General.cpp
```diff
@@ -181,3 +181,25 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
...
+ */
+HandlerResponse WSRequestHandler::HandleGetVideoInfo(WSRequestHandler* req) {
+	obs_video_info ovi; obs_get_video_info(&ovi);
+	OBSDataAutoRelease response = obs_data_create();
+	obs_data_set_int(response, "base_width", ovi.base_width);
+	obs_data_set_int(response, "base_height", ovi.base_height);
+	obs_data_set_int(response, "output_width", ovi.output_width);
+	obs_data_set_int(response, "output_height", ovi.output_height);
+	obs_data_set_double(response, "fps", (double)ovi.fps_num / ovi.fps_den); //TODO: Convert to floating-point FPS?
+	return req->SendOKResponse(response);
```
Palakis: Would be nice to provide properties for output format, YUV range + colorspace (if YUV) and scale filter type.

src/WSRequestHandler_General.cpp
```diff
@@ -181,3 +181,25 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
...
+ * @since 4.6.0 
+ */
+HandlerResponse WSRequestHandler::HandleGetVideoInfo(WSRequestHandler* req) {
+	obs_video_info ovi; obs_get_video_info(&ovi);
+	OBSDataAutoRelease response = obs_data_create();
+	obs_data_set_int(response, "base_width", ovi.base_width);
+	obs_data_set_int(response, "base_height", ovi.base_height);
+	obs_data_set_int(response, "output_width", ovi.output_width);
+	obs_data_set_int(response, "output_height", ovi.output_height);
+	obs_data_set_double(response, "fps", (double)ovi.fps_num / ovi.fps_den); //TODO: Convert to floating-point FPS?
```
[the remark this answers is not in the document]
Rosuav: Oops, I left the TODO on that even after making the suggested change. (Previous iterations of this simply exposed fps_num and fps_den directly.) It is already exposing it as a floating-point FPS value.

Rosuav:
Adjusted per comments/recommendations.

I'm not sure what the difference is between "Number" and "double" in the documentation. It appears that integers use "Number" but doubles use "double"? But sometimes integers use "int". The JSON data type is the same for all of them anyway.

Couldn't find a good example of rendering an enumerated value, so I went with a simple switch statement (with a fallback to integer in case new options get added to OBS).

Palakis:
@Rosuav Thanks, gonna review this.

> Couldn't find a good example of rendering an enumerated value, so I went with a simple switch statement 

I do this as well in several places of the source code so it's good as-is.

Palakis:
@Rosuav 
> I'm not sure what the difference is between "Number" and "double" in the documentation. It appears that integers use "Number" but doubles use "double"? But sometimes integers use "int". The JSON data type is the same for all of them anyway.

Integers should use `int` and floats/doubles use `double` in the docs, because it's the two number types supported by `obs_data_t`. Gonna remove uses of `Number` everywhere else because they're too ambiguous.

Rosuav:
Perfect, that answers that question. Going with `int` then.

src/WSRequestHandler_General.cpp
```diff
@@ -181,3 +181,84 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
...
+	obs_video_info ovi;
+	obs_get_video_info(&ovi);
+	OBSDataAutoRelease response = obs_data_create();
+	obs_data_set_int(response, "baseWidth", ovi.base_width);
+	obs_data_set_int(response, "baseHeight", ovi.base_height);
+	obs_data_set_int(response, "outputWidth", ovi.output_width);
+	obs_data_set_int(response, "outputHeight", ovi.output_height);
+	obs_data_set_double(response, "fps", (double)ovi.fps_num / ovi.fps_den);
+
+	switch (ovi.output_format) {
```
Palakis: Can you put the four `switch` statements in separate functions that return the string value please? Makes them reusable and keeps the main function clutter-free.

src/WSRequestHandler_General.cpp
```diff
@@ -181,3 +181,84 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
...
+		#define CASE(x) case x: obs_data_set_string(response, "videoFormat", #x); break;
```
Palakis: Good idea!

src/WSRequestHandler_General.cpp
Palakis: You can put them in the same file, after include statements & defines and before member functions.

Rosuav: Hmm. They currently don't just return a string - the fallback case has them set an integer. If they get refactored into those convenient functions, what should the fallback do? Use a static buffer to represent the integer? Return NULL and let the caller cope?

Palakis: Use the none/unknown enum value as the fallback value. e.g `VIDEO_FORMAT_NONE`

Rosuav: Done!

Rosuav:
Awesome! Thank you!

PR #327, opened 2019-05-19
Update protocol.md

DieserMerlin:
Removed "version" response field from "GetVersion" request since the websocket doesn't return it any longer


Palakis:
@DieserMerlin Keeping this field for now. Added the field back in https://github.com/Palakis/obs-websocket/commit/0d8999d64f39252ce4165fe9f6fb744ba03dafab

PR #338, opened 2019-06-01
Create a command for opening a projector.

### LINKED ISSUE #214: Request a projector window?
##### Issue type
Feature request

##### Description
I'd love to be able to signal OBS to open a windowed projector. Once the window is open, I can script its position etc using external tools, but currently it requires a mouse click to open the projector itself.

##### Technical information
- **Operating System** : Debian GNU/Linux
- **OBS Studio version** : 19.0.3+ built from git

Rosuav:
Depends on obs-studio#1910 for back end functionality.

Is there a convenient and safe way to have this return a tidy error message if OBS doesn't support obs_frontend_open_projector()?

Closes #214

### DIFF
--- a/src/WSRequestHandler_General.cpp
+++ b/src/WSRequestHandler_General.cpp
@@ -300,7 +300,7 @@ HandlerResponse WSRequestHandler::HandleGetVideoInfo(WSRequestHandler* req) {
 }
 
 /**
- * Open a projector window or create a projector on a monitor.
+ * Open a projector window or create a projector on a monitor. Requires OBS v24.0.4 or newer.
  * 
  * @param {String (Optional)} `type` Type of projector: Preview (default), Source, Scene, StudioProgram, or Multiview (case insensitive).
  * @param {int (Optional)} `monitor` Monitor to open the projector on. If -1 or omitted, opens a window.
@@ -313,7 +313,7 @@ HandlerResponse WSRequestHandler::HandleGetVideoInfo(WSRequestHandler* req) {
  * @since 4.7.0 
  */
 HandlerResponse WSRequestHandler::HandleOpenProjector(WSRequestHandler* req) {
-	#if LIBOBS_API_VER >= 0x18000003
+	#if LIBOBS_API_VER >= 0x18000004
 	const char *type = obs_data_get_string(req->data, "type");
 	int monitor = -1;
 	if (req->hasField("monitor")) {
@@ -324,6 +324,6 @@ HandlerResponse WSRequestHandler::HandleOpenProjector(WSRequestHandler* req) {
 	obs_frontend_open_projector(type, monitor, geometry, name);
 	return req->SendOKResponse();
 	#else
-	return req->SendErrorResponse("Projector opening requires libobs v21.0.3 or newer.");
+	return req->SendErrorResponse("Projector opening requires libobs v21.0.4 or newer.");
 	#endif
 }
--- a/src/WSRequestHandler_General.cpp
+++ b/src/WSRequestHandler_General.cpp
@@ -310,13 +310,13 @@ RpcResponse WSRequestHandler::GetVideoInfo(const RpcRequest& request) {
  * 
  * @param {String (Optional)} `type` Type of projector: Preview (default), Source, Scene, StudioProgram, or Multiview (case insensitive).
  * @param {int (Optional)} `monitor` Monitor to open the projector on. If -1 or omitted, opens a window.
- * @param {String (Optional)} `geometry` Size and position of the projector window (only if monitor is -1). Encoded in base 64.
+ * @param {String (Optional)} `geometry` Size and position of the projector window (only if monitor is -1). Encoded in base 64. Corresponds to OBS's saved projectors.
  * @param {String (Optional)} `name` Name of the source or scene to be displayed (ignored for other projector types).
  * 
  * @api requests
  * @name OpenProjector
  * @category general
- * @since unreleased 
+ * @since unreleased
  */
 RpcResponse WSRequestHandler::OpenProjector(const RpcRequest& request) {
 #if LIBOBS_API_VER >= 0x18000004
--- a/src/WSRequestHandler_General.cpp
+++ b/src/WSRequestHandler_General.cpp
@@ -316,10 +316,10 @@ RpcResponse WSRequestHandler::GetVideoInfo(const RpcRequest& request) {
  * @api requests
  * @name OpenProjector
  * @category general
- * @since 4.7.0 
+ * @since unreleased 
  */
 RpcResponse WSRequestHandler::OpenProjector(const RpcRequest& request) {
-	#if LIBOBS_API_VER >= 0x18000004
+#if LIBOBS_API_VER >= 0x18000004
 	const char *type = obs_data_get_string(request.parameters(), "type");
 	int monitor = -1;
 	if (request.hasField("monitor")) {
@@ -329,7 +329,7 @@ RpcResponse WSRequestHandler::OpenProjector(const RpcRequest& request) {
 	const char *name = obs_data_get_string(request.parameters(), "name");
 	obs_frontend_open_projector(type, monitor, geometry, name);
 	return request.success();
-	#else
+#else
 	return request.failed("Projector opening requires libobs v24.0.4 or newer.");
-	#endif
+#endif
 }
--- a/src/WSRequestHandler_General.cpp
+++ b/src/WSRequestHandler_General.cpp
@@ -324,6 +324,6 @@ HandlerResponse WSRequestHandler::HandleOpenProjector(WSRequestHandler* req) {
 	obs_frontend_open_projector(type, monitor, geometry, name);
 	return req->SendOKResponse();
 	#else
-	return req->SendErrorResponse("Projector opening requires libobs v21.0.4 or newer.");
+	return req->SendErrorResponse("Projector opening requires libobs v24.0.4 or newer.");
 	#endif
 }
--- a/src/WSRequestHandler_General.cpp
+++ b/src/WSRequestHandler_General.cpp
@@ -298,3 +298,28 @@ HandlerResponse WSRequestHandler::HandleGetVideoInfo(WSRequestHandler* req) {
 	obs_data_set_string(response, "scaleType", describe_scale_type(ovi.scale_type));
 	return req->SendOKResponse(response);
 }
+
+/**
+ * Open a projector window or create a projector on a monitor.
+ * 
+ * @param {String (Optional)} `type` Type of projector: Preview (default), Source, Scene, StudioProgram, or Multiview (case insensitive).
+ * @param {int (Optional)} `monitor` Monitor to open the projector on. If -1 or omitted, opens a window.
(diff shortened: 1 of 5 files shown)


Rosuav:
See https://github.com/obsproject/obs-studio/pull/1910 for prerequisite back end functionality.

DeeJayMX:
Do this function will be added ?

Rosuav:
It can't happen in OBS-WebSocket without first happening in OBS-Studio, which is why this PR is flagged as a WIP. But if you want to build both OBS and the plugin from source, you can do so by manually grabbing these commits.

Rosuav:
Rebased onto 4.x-current. Still requires OBS-Studio change though so this is still pending.

Rosuav:
All three of these PRs have been rebased onto the latest 4.x-current and have been tested (at least briefly) on OBS-Studio 24.0.1.

Rosuav:
The upstream requirement is now in the master branch, but hasn't yet been tagged. Note that there's no easy way to probe for this feature other than a version check, and v24.0.3 doesn't always have the feature, so to be conservative, I'm blocking it until v24.0.4 is tagged. That means that it won't actually be included currently, even though a master build of obs-studio WILL work. Once 24.0.4 exists, this feature should start being used.

Rosuav:
It turns out that OBS skipped 24.0.4 and went straight from 24.0.3 to 24.0.5, but the code still works. This patch is now ready to be merged.

DeeJayMX:
Yeah !

src/WSRequestHandler_General.cpp
```diff
@@ -304,3 +304,32 @@ RpcResponse WSRequestHandler::GetVideoInfo(const RpcRequest& request) {
...
+ * 
+ * @param {String (Optional)} `type` Type of projector: Preview (default), Source, Scene, StudioProgram, or Multiview (case insensitive).
+ * @param {int (Optional)} `monitor` Monitor to open the projector on. If -1 or omitted, opens a window.
+ * @param {String (Optional)} `geometry` Size and position of the projector window (only if monitor is -1). Encoded in base 64.
+ * @param {String (Optional)} `name` Name of the source or scene to be displayed (ignored for other projector types).
+ * 
+ * @api requests
+ * @name OpenProjector
+ * @category general
+ * @since 4.7.0
```
Palakis:
```suggestion
 * @since unreleased 
```

src/WSRequestHandler_General.cpp
```diff
@@ -304,3 +304,32 @@ RpcResponse WSRequestHandler::GetVideoInfo(const RpcRequest& request) {
...
+ * @param {String (Optional)} `geometry` Size and position of the projector window (only if monitor is -1). Encoded in base 64.
+ * @param {String (Optional)} `name` Name of the source or scene to be displayed (ignored for other projector types).
+ * 
+ * @api requests
+ * @name OpenProjector
+ * @category general
+ * @since 4.7.0 
+ */
+RpcResponse WSRequestHandler::OpenProjector(const RpcRequest& request) {
+	#if LIBOBS_API_VER >= 0x18000004
```
Palakis: Code nitpick: don't indent preprocessor statements
```suggestion
#if LIBOBS_API_VER >= 0x18000004
```

src/WSRequestHandler_General.cpp
```diff
@@ -304,3 +304,32 @@ RpcResponse WSRequestHandler::GetVideoInfo(const RpcRequest& request) {
...
+	const char *type = obs_data_get_string(request.parameters(), "type");
+	int monitor = -1;
+	if (request.hasField("monitor")) {
+		monitor = obs_data_get_int(request.parameters(), "monitor");
+	}
+	const char *geometry = obs_data_get_string(request.parameters(), "geometry");
+	const char *name = obs_data_get_string(request.parameters(), "name");
+	obs_frontend_open_projector(type, monitor, geometry, name);
+	return request.success();
+	#else
```
Palakis:
```suggestion
#else
```

src/WSRequestHandler_General.cpp
```diff
@@ -304,3 +304,32 @@ RpcResponse WSRequestHandler::GetVideoInfo(const RpcRequest& request) {
...
+	return request.failed("Projector opening requires libobs v24.0.4 or newer.");
+	#endif
```
Palakis:
```suggestion
#endif
```

src/WSRequestHandler_General.cpp
```diff
@@ -304,3 +304,32 @@ RpcResponse WSRequestHandler::GetVideoInfo(const RpcRequest& request) {
 
 	return request.success(response);
 }
+
+/**
+ * Open a projector window or create a projector on a monitor. Requires OBS v24.0.4 or newer.
+ * 
+ * @param {String (Optional)} `type` Type of projector: Preview (default), Source, Scene, StudioProgram, or Multiview (case insensitive).
+ * @param {int (Optional)} `monitor` Monitor to open the projector on. If -1 or omitted, opens a window.
+ * @param {String (Optional)} `geometry` Size and position of the projector window (only if monitor is -1). Encoded in base 64.
```
Palakis: Could you please add a bit more info on the format of the data before it gets encoded in Base64, please? So that users get to know how to generate the geometry string.

Rosuav: It seems that the geometry string is based entirely on Qt.

https://doc.qt.io/qt-5/qwidget.html#saveGeometry

I'm inclined to not try to document the actual format, but just to report that this ties in with the saved projectors in OBS's own configs. Maybe best would be to have a way to query all current projectors?

Palakis: Alright then. I'll add a mention to Qt's documentation and merge this.

Rosuav:
LGTM! Thank you.

ethaniel:
Thank you so much guys, we need this feature too! When will there be a next release of the plugin? Can’t wait to update :)

ethaniel:
I just tested this in the test build and noticed a feature, where if you call "OpenProjector" several times in full screen mode, then you get multiple full screen windows. I'm not sure if this is a feature of OBS or your code, but just wanted to bring to your attention.

Also, I couldn't find the CloseProjector function. Is there one?

Rosuav:
Yes, you can open multiple FS projectors in the OBS UI too.

CloseProjector isn't currently a thing. If you want that, I'd recommend going for a full set of three (or four):

- ListProjectors
- QueryProjector
- CloseProjector
- And optionally SetProjector or UpdateProjector

Projectors would need some sort of identifier, which would be returned by OpenProjector. You could then list all projectors, and query a single projector. The info you'd get back would include the same four pieces of information that OpenProjector (including the geometry, which would resolve the issue of "what does the geometry mean?"). You'd then be able to close a projector by giving its ID.

I don't know how useful this would be, but if any of it is, it's probably worth making the full set (maybe without any sort of update). If you make it, I'd be happy to help test it.

bobvandevijver:
I would most certainly be interested in a CloseProjector method. Is anyone currently working on the functionality as proposed by @Rosuav?

tt2468:
@bobvandevijver Unfortunately it would not be possible at the moment to create more projector-related requests, as the only available api endpoint in OBS for projectors is `obs_frontend_open_projector`

bobvandevijver:
Well, that endpoint was added by @Rosuav to OBS (https://github.com/obsproject/obs-studio/pull/1910), so it should probably be possible to do the same for the proposed methods.

tt2468:
Ah I didnt notice, ok

Rosuav:
The ability to close projectors would depend on some way of identifying them. So there'd probably need to be a "list projectors" API, and some sort of unique ID, and validation of those IDs. (And then open_projector could return that ID.) And yeah, it'd all need to start with having that functionality in OBS itself. Feel free to open a tracker issue there, state your use case etc, and tag me (not that I'm promising to implement it, though I may end up doing so).

bobvandevijver:
They only allow issues for bugs in the OBS repository, suggestions such as these should be posted to their ideas (https://ideas.obsproject.com/) page... Can't really tag you over there I guess 🤷‍♂️

bobvandevijver:
So, I struggled with the geometry parameter, but I figured it out and wanted to share this here. The latest OBS only supports version 2 of the saveGeometry call, which means it is Qt < 5.12. 

For the `saveGeometry` source you can check [this](https://code.woboq.org/qt5/qtbase/src/widgets/kernel/qwidget.cpp.html#_ZNK7QWidget12saveGeometryEv) page.

This [base64 tool](https://cryptii.com/pipes/base64-to-hex) helped me during debugging.

I used the following Qt code to create this geometry. Can be used with any Qt example, as long as you have a widget (`form` here).

```cpp 
            form.show();
            form.setGeometry(320, 190, 1280, 720);
            QByteArray geometry = form.saveGeometry();
            // Version 3
            qDebug() << QString(geometry.toBase64());
            // Convert to version 2
            geometry.remove(geometry.length() - 16, 16);
            geometry[5] = 2;
            qDebug() << QString(geometry.toBase64());
```

Note that the geometry implementation can be a bit [unpredictable](https://doc.qt.io/qt-5/application-windows.html#window-geometry)...

foxxyz:
@bobvandevijver which OS did you test on?

I tried your geometry string on MacOS and didn't see any difference between omitting it or including it. In fact, I wasn't able to get OBS to place the projector window in any other place but the default.

Also, decoding the string shows `x = 320`, `y = 190`, `width = 1600`, `height = 910`. Correct to assume you created the geometry with those values, and not with `320, 190, 1280, 720`?

bobvandevijver:
@foxxyz I used Windows 10 (x64), I never tested on MacOS. But, the string value and the values I posted are exactly as executed on my machine. 

As already stated, the geometry implementation of Qt is a bit unpredictable (or maybe even better, weird), but we can clearly see in the source that the Rect that is being encoded defines the screen points of the window, so `x1`, `x2`, `y1` & `y2`, instead of the origin `x` & `y` and the `height` & `width` (see the source, more specifically the [`frameGeometry`](https://code.woboq.org/qt5/qtbase/src/widgets/kernel/qwidget.cpp.html#_ZNK7QWidget13frameGeometryEv) call).

However, it also includes the geometry and state of the screen you're using (see the [`normalGeometry`](https://code.woboq.org/qt5/qtbase/src/widgets/kernel/qwidget.cpp.html#_ZNK7QWidget14normalGeometryEv) call) and the screen number. That might be why it doesn't work for you with this specific string.

I would recommend that you install the Qt developer environment, and execute my suggested code yourself: you will probably get another result. I believe the main reason for this is that the string representation is not meant for portability, and it just guaranteed to work on the same machine with the same screen setup. That's probably also why it is just ignored when Qt believes it to be invalid for some reason: I struggled for quite some time to actually get a working string that would be picked up correctly...

The better solution would be to change the actual OBS API for the projector method, by either removing the `geometry` and replace it with the `x` & `y` and the `height` & `width`, or retrieve the current position through the new API call @Rosuav suggested earlier in this PR. But both require work on the OBS side.

foxxyz:
Thank you for the detailed response! This helps a lot!

Agreed that the OBS API should probably just take in `x`, `y`, `width` and `height`, especially if there are OS differences in Qt geometry output.

PR #354, opened 2019-07-21
Make it possible to broadcast messages (data) to other websocket clients

julijane:
This PR adds a new function "BroadcastWebSocketMessage" which takes two parameters: realm and data. realm is intended to be a user-defined string to distinguish the source/recipient/scope of the message and data is the actual message.

Why?

With this change for example a javascript code in a browser source can connect to the websocket server and can receive such Messages and act upon them. This makes it possible for example to write some remote controlling software which triggers actions in a browser source without the need for any additional server.

### DIFF
--- a/src/WSEvents.cpp
+++ b/src/WSEvents.cpp
@@ -1506,6 +1506,25 @@ void WSEvents::OnStudioModeSwitched(bool checked) {
 }
 
 /**
+ * A custom broadcast message was received
+ *
+ * @param {String} `realm` Identifier provided by the sender
+ * @param {Object} `data` User-defined data
+ *
+ * @api events
+ * @name BroadcastCustomMessage
+ * @category general
+ * @since 4.7.0
+ */
+void WSEvents::OnBroadcastCustomMessage(QString realm, OBSDataAutoRelease data) {
+	OBSDataAutoRelease broadcastData = obs_data_create();
+	obs_data_set_string(broadcastData, "realm", realm.toUtf8().constData());
+	obs_data_set_obj(broadcastData, "data", data);
+
+	broadcastUpdate("BroadcastCustomMessage", broadcastData);
+}
+
+/**
  * @typedef {Object} `OBSStats`
  * @property {double} `fps` Current framerate.
  * @property {int} `render-total-frames` Number of frames rendered
--- a/src/WSEvents.cpp
+++ b/src/WSEvents.cpp
@@ -1516,7 +1516,7 @@ void WSEvents::OnStudioModeSwitched(bool checked) {
  * @category general
  * @since 4.7.0
  */
-void WSEvents::OnBroadcastCustomMessage(QString realm, OBSDataAutoRelease data) {
+void WSEvents::OnBroadcastCustomMessage(QString realm, obs_data_t* data) {
 	OBSDataAutoRelease broadcastData = obs_data_create();
 	obs_data_set_string(broadcastData, "realm", realm.toUtf8().constData());
 	obs_data_set_obj(broadcastData, "data", data);
--- a/src/WSRequestHandler_General.cpp
+++ b/src/WSRequestHandler_General.cpp
@@ -232,6 +232,45 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
 }
 
 /**
+ * Broadcast some Data (String) to all connected Websocket-Clients
+ *
+ * @param {String} `realm` Some Identifier to be choosen by the client
+ * @param {String} `data` User-defined data String
+ *
+ * @api requests
+ * @name Authenticate
+ * @category general
+ * @since 0.3
+ */
+HandlerResponse WSRequestHandler::HandleBroadcastWebSocketMessage(WSRequestHandler* req) {
+	if (!req->hasField("realm") || !req->hasField("data")) {
+		return req->SendErrorResponse("missing request parameters");
+	}
+
+	QString realm = obs_data_get_string(req->data, "realm");
+	QString data = obs_data_get_string(req->data, "data");
+
+	if (realm.isEmpty()) {
+		return req->SendErrorResponse("realm not specified!");
+	}
+
+	if (data.isEmpty()) {
+		return req->SendErrorResponse("data not specified!");
+	}
+
+	auto events = GetEventsSystem();
+
+	OBSDataAutoRelease broadcastData = obs_data_create();
+	obs_data_set_string(broadcastData, "realm", realm.toUtf8().constData());
+	obs_data_set_string(broadcastData, "data", data.toUtf8().constData());
+
+	events->broadcastUpdate("BroadcastWebSocketMessage", broadcastData);
+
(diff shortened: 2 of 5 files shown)


src/WSRequestHandler_General.cpp
```diff
@@ -231,6 +231,45 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
...
+/**
+ * Broadcast some Data (String) to all connected Websocket-Clients
+ *
+ * @param {String} `realm` Some Identifier to be choosen by the client
+ * @param {String} `data` User-defined data String
+ *
+ * @api requests
+ * @name Authenticate
+ * @category general
+ * @since 0.3
```
Palakis:
```suggestion
 * @since 4.7.0
```

src/WSRequestHandler_General.cpp
```diff
@@ -231,6 +231,45 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
...
 }
 
+/**
+ * Broadcast some Data (String) to all connected Websocket-Clients
+ *
+ * @param {String} `realm` Some Identifier to be choosen by the client
+ * @param {String} `data` User-defined data String
+ *
+ * @api requests
+ * @name Authenticate
```
Palakis:
```suggestion
 * @name BroadcastWebSocketMessage
```

src/WSRequestHandler_General.cpp
```diff
@@ -231,6 +231,45 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
 	return req->SendOKResponse(response);
 }
 
+/**
+ * Broadcast some Data (String) to all connected Websocket-Clients
+ *
+ * @param {String} `realm` Some Identifier to be choosen by the client
+ * @param {String} `data` User-defined data String
+ *
+ * @api requests
```
Palakis:
```suggestion
 * @api general
```

src/WSRequestHandler_General.cpp
```diff
@@ -231,6 +231,45 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
 	return req->SendOKResponse(response);
 }
 
+/**
+ * Broadcast some Data (String) to all connected Websocket-Clients
+ *
+ * @param {String} `realm` Some Identifier to be choosen by the client
+ * @param {String} `data` User-defined data String
```
Palakis: It'd be better to make it an object instead.

```suggestion
 * @param {Object} `data` User-defined data object
```

julijane:
@Palakis Thank you for your review. I made the requested changes.

julijane:
@Palakis I was thinking about this. I think it would be neat if the broadcast would contain a identifier (e.g. the string from getRemoteEndpoint) for the source and the possibility to then send a response to just this source (so another request SendWebSocketMessage which gets realm, data and target). However to do this the source string (or the connection_hdl) would need to be available in the RequestHandler. This should be doable but what are your feelings about this idea?

src/WSRequestHandler_General.cpp
```diff
@@ -231,6 +231,45 @@ HandlerResponse WSRequestHandler::HandleGetStats(WSRequestHandler* req) {
...
+ *
+ * @param {String} `realm` Some Identifier to be choosen by the client
+ * @param {Object} `data` User-defined data
+ *
+ * @api general
+ * @name BroadcastWebSocketMessage
+ * @category general
+ * @since 4.7.0
+ */
+HandlerResponse WSRequestHandler::HandleBroadcastWebSocketMessage(WSRequestHandler* req) {
```
Palakis: Can you also please add at the top of this file the specification for the associated event, please?

Palakis: personal nitpick: I believe `BroadcastCustomMessage` for the request type and `OnCustomMessage` for the event would be more fitting names

julijane: @Palakis I changed it to BroadcastCustomMessage and CustomMessage (the other events also don't have the "On"-Prefix, only in the name of the function). I don't understand the part about "associated event", I checked the source in order to understand what you might mean but did not find out. Please advise.

Palakis: > I don't understand the part about "associated event", I checked the source in order to understand what you might mean but did not find out. Please advise.

The `CustomMessage` event must be documented like the events in `WSEvents.cpp`. Events spec example: https://github.com/Palakis/obs-websocket/blob/4.x-current/src/WSEvents.cpp#L495-L504

You can either leave the `CustomMessage` event here and document it at the top of this file, or move it into a `WSEvents::OnCustomMessage` method in `WSEvents.cpp` (with `BroadcastCustomMessage` calling it from here) and document it there.

Palakis: The spec for `CustomMessage` would look something like this:

```
/**
 * Custom message from another client.
 *
 * @return {String} `realm` User-defined message identifier
 * @return {Object} `data` User-defined data
 *
 * @api events
 * @name CustomMessage
 * @category general
 * @since 4.7.0
 */
```

julijane: I have splitted it.

julijane: I used the name OnBroadcastCustomMessage still because as suggested before it might be nice to also have custom messages which are not broadcasted but directed at a target.

Palakis: Agreed! :+1:

src/WSEvents.cpp
```diff
@@ -1505,6 +1505,25 @@ void WSEvents::OnStudioModeSwitched(bool checked) {
...
+ *
+ * @param {String} `realm` Identifier provided by the sender
+ * @param {Object} `data` User-defined data
+ *
+ * @api events
+ * @name BroadcastCustomMessage
+ * @category general
+ * @since 4.7.0
+ */
+void WSEvents::OnBroadcastCustomMessage(QString realm, OBSDataAutoRelease data) {
```
Palakis: `OBSDataAutoRelease` doesn't increase the reference count but only decreases it. As the calling `WSRequestHandler::BroadcastCustomMessage` function also uses `OBSDataAutoRelease`, this may cause a refcount error.

Either increase it as well by using `OBSData`, or use a plain `obs_data_t*` to not care with reference counting at all (and leave that responsibility to its caller).

julijane: done

julijane:
@Palakis Something went wrong with the automatic protocol.md generation. The entry in event section is missing the return values and the entry in the command section is missing alltogether.

Palakis:
@julijane Both the command spec used `@param` (specific to typedefs) instead of `@return`, and the command spec had its `@api` field set to `general` instead of `requests`. Just fixed it.

julijane:
@Palakis Oops i'm sorry for that mistake. Thanks for fixing already!

julijane:
@Palakis You made a mistake though, see PR

PR #355, opened 2019-07-24
Allow clients to save persistent data on the server

### LINKED ISSUE #184: Server-side datastore
##### Issue type
Feature request (and implementation discussion; I'm happy to do the work)

##### Description
I use multiple endpoints to control my OBS instances via obs-websocket--I have a few different tablets available to different people, on camera and off, for flipping between scenes. However, as I start looking to add some configuration and some options to my tooling outside of just modifying the running state of the application it's becoming evident that I need a place to store data that all my clients can access. I can totally do something with Dropbox or some other out-of-band store--but it probably makes sense for it to just be on the box so that any client with the password can connect to it and just be done with it. I kind of just envision a "obs-websocket.json" file in %appdata% or the like where individual tools can define their own top-level key, i.e. mine would be "StreamCopilot", and just provide something like a [jsondiffpatch](https://github.com/benjamine/jsondiffpatch) or even a jsonpath to setting a particular key with a value. That way, a client just has to connect to OBS and all of the dedicated programmed settings for their hotkeys, etc. work on every client they want to use. I'm sure other folks using obs-websocket probably have a use for it too.

I am happy to do the legwork to implement this feature if we can nail down what it should look like, as this would be _super_ helpful to me and I'd love to contribute.

julijane:
This PR adds two WebSocket functions which allow the user to store and retrieve arbitrary data on the server. The data will be saved in the module-config directory for obs-websocket. Filename is built from the user provided realm but sanitized.

Why?
This would allow applications using obs-websocket to store application specific data (e.g. config) on the server which can then be shared between different instances of that application.

_Edit by Palakis:_
Closes #184


julijane:
I have added a size limitation of 32 KiB for the stored data. It is a bit of a moot point though, as the user could add as many 32 KiB chunks as he wants. Maybe obs-websocket needs some sort of permission system and the possibility of having multiple credentials with different permissions?

julijane:
I have added configuration elements for enabling/disabling persistent storage and configuration of maximum size per saved scope/realm.

Palakis:
@julijane Thanks for your work on this! It is a neat feature that was requested by several users and your PR is good.

Regarding how data is stored, I'm not exactly a fan of using several files for each realm. I'd be more comfortable with storing everything into the same spot: a QJsonObject instance that gets serialized on load and deserialized on save into a global config key. This would be closer to the Key-Value store concept I had in mind for this feature, and realms would still be there as root elements of the QJsonObject.

What do you think about this?

P.S. : JSONPath could be a cool addition as well, but I just putting this as an idea here. That's not required for this PR and can be added later.

julijane:
@Palakis Please review the updated PR.

I have adopted they "key-value" idea and have changed it as suggested i.e. it now stores the data in a config key.

src/Config.h
```diff
@@ -42,14 +42,17 @@ class Config {
...
 
 		bool DebugEnabled;
 		bool AlertsEnabled;
+		bool PersistentDataEnabled;
 
 		bool AuthRequired;
 		QString Secret;
 		QString Salt;
 		QString SessionChallenge;
+		QString PersistentDataStore;
```
Palakis, line 55: In-memory, I'd rather have the data store be a QJsonDocument, and have it serialized on config save and deserialized on config load. Deserializing and serializing the whole store on every change (request) has a huge performance cost.

src/WSRequestHandler_General.cpp
```diff
@@ -281,6 +287,102 @@ RpcResponse WSRequestHandler::BroadcastCustomMessage(const RpcRequest& request)
...
+	if (value.isEmpty()) {
+		return request.failed("value not specified!");
+	}
+
+	if (value.length() > GetConfig()->PersistentDataMaxSize) {
+		return request.failed("maximum data size exceeded, check configuration");
+	}
+
+	QString jsonString = GetConfig()->PersistentDataStore;
+	QJsonDocument jsonDoc = QJsonDocument::fromJson(jsonString.toUtf8());
```
Palakis, line 326: Requests can be processed on separate threads (even Get/SetPersistentData). It'd be best to throw a mutex in there to protect the data store from conflicts between two requests (e.g. two changes in parallel with the last discarding the first)

Palakis (changes requested):
Looks good overall! There's one more thing we need to address and we'll be good to go.

src/Config.h
julijane: Well this is what you requested before :)

> Regarding how data is stored, I'm not exactly a fan of using several files for each realm. I'd be more comfortable with storing everything into the same spot: a QJsonObject instance that gets serialized on load and deserialized on save into a global config key. 

I'm unsure how to do what you ask for now. If I don't serialize on change then the change would not persist when for example OBS crashes or am I missing something?

Palakis: I think the confusion lies here: "deserialized on load and serialized on save" (opposite of what I wrote btw). To be more precised, it's deserialized on config load and serialized on config save, so it exists as a QJsonObject in the Config object.

src/WSRequestHandler_General.cpp
julijane: You want to keep it in-memory as a QJsonbject. According to the documentation all functions of QJsonobject are reentrant.

https://doc.qt.io/qt-5/qjsondocument.html

> can also be called simultaneously from multiple threads, but only if each invocation uses its own data

So it should only fail if two clients write to the same key then and this I would perceive as more of a use error as even with a mutex this would result in unpredictable / discarded changes.

src/Config.h
julijane: It would be better to use QJsonObject in the Config Object, it is created from the QJSonDocument on the fly and would break reentrancy (see other comment). I do not see any downside of directly putting the QJsonObject in the config object?

src/WSRequestHandler_General.cpp
Palakis: From what I understand here, QJsonDocument's methods are reetrant because they don't rely on global/static data to function, apart from the JSON data owned by a QJsonDocument instance. However, the data structure used for JSON data in QJsonDocument is not guaranteed to be thread-safe, hence the need to protect it with a mutex.

tt2468:
Will be released in a different form in the 5.x series using more effective data storage techniques.

julijane:
@tt2468 That's good to hear. I unfortunately never found the time to work on this and fulfill the requests by Palakis, so it is good to hear that someone else is taking this over and if you do it in way you find is favorable that is even better. Thank you.

tt2468:
Yes, the plan is to store persistent data as dedicated .json files, and different requests for storing those profiles in both global and per-profile datastores. There will also likely be globally cached JSON objects to improve performance.

julijane:
@tt2468 Now that is a bit funny though, as my initial attempt actually did put it into separate files and then Palakis requested me to change that :)

> Regarding how data is stored, I'm not exactly a fan of using several files for each realm. I'd be more comfortable with storing everything into the same spot

Not complaining, but pointing this out :)

tt2468:
One file per profile, one file for global store. Regardless of realm

julijane:
@tt2468 Well so the new plan is a mixture of both old viewpoints I guess. But as I said, I am happy that someone is taking this over and working on it. I really think this feature has a lot of potential use cases so it is good that it will now be finally happening. Thank you for all your work on obs-ws.

PR #356, opened 2019-07-26
Expand GetSceneItems with a 'full' flag for more info

Rosuav:
As suggested in #324, it may be better to keep the events lightweight, but have a "more info pls" flag on the explicit GetSceneItems. This keeps the current paths fast and simple, but allows a one-round-trip request that will give additional information.

Currently, this additional information is a `muted` flag on audio sources, and the `alignment` (gravity) on visual ones.

### DIFF
--- a/docs/generated/comments.json
+++ b/docs/generated/comments.json
@@ -17,7 +17,9 @@
         "{Number} `x`",
         "{Number} `y`",
         "{String (optional)} `parentGroupName` Name of the item's parent (if this item belongs to a group)",
-        "{Array<SceneItem> (optional)} `groupChildren` List of children (if this item is a group)"
+        "{Array<SceneItem> (optional)} `groupChildren` List of children (if this item is a group)",
+        "{Boolean (optional)} `muted` Whether the source has been muted (only included if full info was requested)",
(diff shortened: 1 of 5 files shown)


Rosuav:
Rebased onto 4.x-current with no changes.

Palakis:
@Rosuav I get the idea for this, but do you currently have a use for this in `GetCurrentScene` ?

I have the feeling we're kinda exposing internal logic that shouldn't be exposed, even though I agree we should keep events lightweight. IMHO, `GetCurrentScene` should internally have full always set to true, with full set to false for events.

Or maybe the event code should just pick whatever fields it wants from the data object returned by `Utils::GetSceneItems`.

Rosuav:
It's for my increasingly-inaccurately-named volume control: https://github.com/Rosuav/obs-remote-volume

In response to a scene change event or when explicitly requested with GetCurrentScene, it updates its displayed state:

https://github.com/Rosuav/obs-remote-volume/blob/master/index.js#L242

The element's position (source.x, source.y) is sent, but without the alignment, we can't actually draw it on screen. In order to use the x and y coordinates, I would have to fire off a separate request for every element, just for that one piece of information. Otherwise, the x and y are basically useless.

Rosuav:
All three of these PRs have been rebased onto the latest 4.x-current and have been tested (at least briefly) on OBS-Studio 24.0.1.

Rosuav:
All the functionality from this PR has now been implemented elsewhere. Closing in favour of #428.

PR #360, opened 2019-08-22
allow for "linuxbrowser_source" to be controlled

ZyanKLee:
based on the request at https://github.com/bazukas/obs-linuxbrowser/issues/46 I did a quick search for the string "browser_source" and these seem to be the only two occurrences. Enhancing the conditional with the identifier of the obs-linuxbrowser plugin, this should solve the problem.

As I'm currently not at home, I can't verify this does what it should, but I will test it later tonight.

@mjc666 if you are still interested please give it a test run to verify if these two plugins work together now.

### DIFF
--- a/src/WSRequestHandler_Sources.cpp
+++ b/src/WSRequestHandler_Sources.cpp
@@ -923,7 +923,7 @@ HandlerResponse WSRequestHandler::HandleGetBrowserSourceProperties(WSRequestHand
 	}
 
 	QString sourceId = obs_source_get_id(source);
-	if (sourceId != "browser_source" && sourceId != "linuxbrowser_source") {
+	if (sourceId != "browser_source" && sourceId != "linuxbrowser-source") {
 		return req->SendErrorResponse("not a browser source");
 	}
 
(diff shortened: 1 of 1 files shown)


ZyanKLee:
I compiled the plugin and installed it locally. I did not have enough time tonight to really try and control the linuxbrowser plugin from the websocket, but at least nothing broke by my change.

ImportTaste:
@ZyanKLee I'm interested in the compiled version of the plugin. The build environment is a lot more than I want to set up right now. Would you be willing to share it? This project is GPL 2.0 so all that would have to be included is the license. https://raw.githubusercontent.com/Palakis/obs-websocket/4.x-current/LICENSE

src/WSRequestHandler_Sources.cpp
```diff
@@ -923,7 +923,7 @@ HandlerResponse WSRequestHandler::HandleGetBrowserSourceProperties(WSRequestHand
 	}
 
 	QString sourceId = obs_source_get_id(source);
-	if (sourceId != "browser_source") {
+	if (sourceId != "browser_source" && sourceId != "linuxbrowser_source") {
```
Palakis: It's `linuxbrowser-source`, not `linuxbrowser_source`: https://github.com/bazukas/obs-linuxbrowser/blob/master/src/plugin/main.c#L630

ZyanKLee:
Thank you, @palakis

PR #376, opened 2019-09-16
HTTP endpoint

Palakis:
Allow HTTP clients to call `POST /execute` to execute commands.

Examples with curl:

```
curl -d '{ "request-type": "GetVersion" }' http://localhost:4444/execute
curl 'http://localhost:4444/execute?request-type=GetVersion'
```

To-do:

- [x] Auth support
- [x] Use an error status code when a request returns an error
- [ ] Query string params support
- [ ] Docs section
- [ ] CORS header


tt2468:
Closing this as it is mainly out of the scope of obs-websocket. Users wanting to query OBS via HTTP request should use https://github.com/IRLToolkit/obs-websocket-http

Issue #383, opened 2019-10-15
TakeSceneScreenshot


BarryCarlyon:
##### Issue type
Feature request

##### Description
Similar to `TakeSourceScreenshot` I'd like a `TakeSceneScreenshot`

I imagine this would only be possible to take a shot of the active scene. Which I don't have a problem with if that was the case

Palakis:
@BarryCarlyon You can already do this with TakeSourceScreenshot. Just specify the scene name in `sourceName`.

PR #384, opened 2019-10-15
Take scene screenshot

BarryCarlyon:
Revise the wording of `TakeSourceScreenshot` to note that if you specify a scene name, you get the scene.

Sorta #383


dodgepong:
This is because scenes are sources. Perhaps the wording should say something like:

    @param {String} `sourceName` Source name. Note that, since scenes are also sources, you can also provide a scene name.

Or perhaps more simply:

    @param {String} `sourceName` Source (or scene) name.

BarryCarlyon:
Yeah that wording is better. Will revise

PR #410, opened 2020-02-07
feat(events): add "TransitionEnd" and "TransitionVideoEnd" events

Lange:
Adds events for `TransitionEnd` and `TransitionVideoEnd`.

`TransitionVideoEnd` is specific to Stinger transitions. It is fired when their video has actually completed playback. This is important, because `TransitionEnd` is fired at the cut point of the Stinger, which is nice to have but isn't always what the user might want.

This builds on #409, and should be merged after it.

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -413,6 +413,30 @@ bool Utils::SetTransitionByName(QString transitionName) {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


Palakis:
@Lange Can you please fix the merge conflict in WSEvents.cpp?

Lange:
@Palakis Done!

Issue #408, opened 2020-02-06
TransitionBegin reports the wrong duration of Stinger Overrides


Lange:
##### Issue type
Bug report 🐛

##### Description
`TransitionBegin` reports the `duration` of the transition incorrectly when the transition is both an Override and a Stinger.

##### Steps to reproduce and other useful info
1. Create a Stinger transition
2. Assign it as a Transition Override on a Scene
3. Transition to that scene
4. Observe that the `duration` reported by `TransitionBegin` will be whatever this numeric input is set to, and not the actual duration of the Stinger: 
![image](https://user-images.githubusercontent.com/873012/73988487-7c2b4d00-4908-11ea-81c9-9f33da21648b.png)

##### Technical information
- **Operating System** : Win 10 x64 Pro 18363
- **OBS Studio version** : 24.0.3

Palakis:
Fixed by #411

PR #411, opened 2020-02-07
fix(events): don't advertise a false duration for fixed transitions

Lange:
Some transitions, such as Stingers, have a fixed duration. There is currently no way to retrieve this value from the API, and `obs-websocket` will currently return an incorrect value when one of these transitions begins.

For now, I think it is better to explicitly return -1, which is a better indicator that we have no idea what the real duration is, instead of a misleading number which users might think looks correct at first glance.

Related to #408

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -387,6 +387,13 @@ int Utils::GetTransitionDuration(obs_source_t* transition) {
 		return 0;
 	}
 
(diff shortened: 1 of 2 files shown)


Palakis:
@Lange Merge conflict here as well

Lange:
@Palakis Done!

PR #416, opened 2020-02-15
Fix SetStreamSettings when switching type

### LINKED ISSUE #378: SetStreamSettings not applying when switching from rtmp_common to rtmp_custom
##### Issue type
Bug report

##### Description
When switching the settings type from rtmp_common to rtmp_custom (along with saving), the server responds as if the settings have changed, but they stay as rtmp_custom.

1. Set the output stream to twitch, then set any stream key
2. With the websocket, set the new type to rtmp_custom, along with a stream server and key
3. The server will respond with the info you provided.
4. Going into the gui, no info has changed.
5. This is both after setting the save bool to true, and also after running the SaveStreamSettings event.

##### Technical information
- **Operating System** : Windows 10 1809
- **OBS Studio version** : 23.2.1

I've even tried populating any combination of variables. Nothing has worked. This problem exists on any of the common services, not just twitch. When the rtmp type is already set to rtmp_custom, the settings will apply just fine. There is also a notable lack of documentation on the "settings" object.It is referenced but no examples or explanations are ever given.

tt2468:
This fixes #378 by applying the stream settings where currently it defines the service and responds with the data of that service object, but does not actually apply it. This also changes the behavior of the response to read the result of the change (as if you called getstreamsettings), which should mean that if the change was not successful, the response object should differ from the request object.

### DIFF
--- a/src/WSRequestHandler_Streaming.cpp
+++ b/src/WSRequestHandler_Streaming.cpp
@@ -212,7 +212,7 @@ RpcResponse WSRequestHandler::SetStreamSettings(const RpcRequest& request) {
 	if (requestedType != nullptr && requestedType != serviceType) {
 		OBSDataAutoRelease hotkeys = obs_hotkeys_save_service(service);
 		service = obs_service_create(
-			STREAM_SERVICE_ID, requestedType.toUtf8(), requestSettings, hotkeys);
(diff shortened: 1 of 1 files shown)


Palakis:
@tt2468 Could you squash your commits please?

PR #428, opened 2020-03-11
Add a 'muted' flag to GetSceneItems

Rosuav:
Now that the alignment/gravity of visual scene items is being transmitted, the only part of #356 still remaining is to add a 'muted' flag to aural scene items. This allows a client to easily see which entities are currently muted when querying with GetSceneItems or seeing the full scene details in SwitchScenes; the status can already be updated by the SourceMuteStateChanged event.

Replaces PR #356.


Rosuav:
Thank you!

PR #439, opened 2020-03-18
T-Bar control

### LINKED ISSUE #399: Add support for controling the T-Bar
##### Issue type
Feature request

##### Description
It would be very handy to have support for the new T-Bar feature that was added with https://github.com/obsproject/obs-studio/commit/a97039db64abd788a675b019bd7f2dd61f69952b

Palakis:
----- Added by tt -----
New requests this adds:
- `ReleaseTBar`- A clean-up request if `release` is overrided to `false` when `SetTBarPosition` is called.
- `SetTBarPosition`- Set the TBar transition position (updates in OBS too!)
- `GetTransitionPosition` - Get the position of the TBar whether or not it is in auto (transition button) or manual (tbar) mode.

----------=----------

Fixes #399


tt2468:
It would be nice to have a `GetTBarPosition` request too.

src/WSRequestHandler_Transitions.cpp
```diff
@@ -120,3 +120,53 @@ RpcResponse WSRequestHandler::GetTransitionDuration(const RpcRequest& request) {
...
+	if (obs_transition_fixed(currentTransition)) {
+		return request.failed("current transition doesn't support t-bar control");
+	}
+
+	double position = obs_data_get_double(request.parameters(), "position");
+	if (position < 0.0 || position > 1.0) {
+		return request.failed("position is out of range");
+	}
+
+	obs_transition_set_manual_time(currentTransition, position);
```
tt2468, line 150: This is a big problem, and the `SetTBarPosition` feature will not be worth adding unless it is fixed.

The problem is that in the OBS UI, when you change the t-bar, it does a whole bunch of important stuff before calling this. See [here](https://github.com/cg2121/obs-studio/blob/68d88fa4c6e28edf2368a38ad8bc1f2403d0231d/UI/window-basic-main-transitions.cpp#L921)

So what actually has to happen is to first call `TBarChanged`, then `TBarReleased` needs to be called to clean up the changes.

PR #449, opened 2020-03-29
Target Scene Items per ID, in addition to scene name and source name

### LINKED ISSUE #154: Unable to distinguish between duplicate items in a scene for all current and future applicable methods
##### Issue type
Feature Request / Brainstorm

##### Description
Duplicate scene items have the same name so it is impossible to change get or set properties of each duplicate other than the first.

Possible implementations:
- Integer indicator for instance of a scene item with the same name
- Integer indicator for position of an item in a scene
- ID of the item (unique and persisting through restart ideally)

@haganbmj thoughts/feeback?

### LINKED ISSUE #398: API: SetSceneItemProperties should accept item.id for handling 1 scene:N items:1 source
##### Issue type
Feature request

##### Description
Many requests provide for a scene item ID field `item.id (int)` in the request, but `GetSceneItemProperties` and `SetSceneItemProperties` do not yet. Instead, currently scene items are addressed by name only, and the API implementation picks the first matching item.

However, **duplicated scene items always have the same name. ID specificity, or a similar feature, is required in order for clients to be able to address duplicated items within the same scene.**

Example use case: Restreaming someone else's stream for a race. Often there is 1 VLC source for the racer's own stream. This stream is used as an input source in the restream's OBS, where it is then cropped into multiple pieces (to pull in the racer's game, racer's timer, racer's webcam...). To achieve multiple crops from a single high bandwidth input, the OBS user duplicates the VLC scene item and crops each copy differently. These items have the same source and scene item name, but different IDs and crops. I would like to read and write each of these crops.

(No Workaround: I considered a workaround where I would reorder the scene items to cause the Properties request to address different items as the "first item". However, I don't see a way to maintain scene item groups when reordering. Much obliged if anyone else can find a workaround for now.)

##### Technical information
- **Operating System** : Windows 10, Linux
- **OBS Studio version** : 24.0.3

Palakis:
I went through the spec of all scene item request types and events. All scene items events are fine. There are three request types to fix:
- [x] [`GetSceneItemProperties`](docs/generated/protocol.md#GetSceneItemProperties)
  - [x] Add `itemId` response field
- [x] [`SetSceneItemProperties`](docs/generated/protocol.md#SetSceneItemProperties)
- [x] [`ResetSceneItem`](docs/generated/protocol.md#ResetSceneItem)

In these three request types, allow the `item` param to either be:
- A string, targeting the first scene item available by this name (current/legacy behaviour)
- An object, borrowing the spec used for the `item` request field in [`DeleteSceneItem`](docs/generated/protocol.md#DeleteSceneItem) and [`DuplicateSceneItem`](docs/generated/protocol.md#DuplicateSceneItem)
   - [x] For consistency, backport new behaviour to DeleteSceneItem and DuplicateSceneItem

Fixes #154
Fixes #398

### DIFF
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -299,13 +299,13 @@ obs_sceneitem_t* Utils::GetSceneItemFromItem(obs_scene_t* scene, obs_data_t* ite
 
 obs_sceneitem_t* Utils::GetSceneItemFromRequestField(obs_scene_t* scene, obs_data_item_t* dataItem)
 {
-	obs_data_type dataType = obs_data_item_gettype(dataItem);
+	enum obs_data_type dataType = obs_data_item_gettype(dataItem);
 
 	if (dataType == OBS_DATA_OBJECT) {
 		OBSDataAutoRelease itemData = obs_data_item_get_obj(dataItem);
 		return GetSceneItemFromItem(scene, itemData);
 	} else if (dataType == OBS_DATA_STRING) {
-		const char* name = obs_data_item_get_string(dataItem);
+		QString name = obs_data_item_get_string(dataItem);
 		return GetSceneItemFromName(scene, name);
 	}
 
--- a/src/Utils.cpp
+++ b/src/Utils.cpp
@@ -180,34 +180,6 @@ obs_data_t* Utils::GetSceneItemData(obs_sceneitem_t* item) {
 	return data;
 }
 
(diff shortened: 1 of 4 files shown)


alfalah919:
alfalah

alfalah919:
#@login!9)/+#!;;!!$(

PR #469, opened 2020-04-21
Add simpleobsws to list of language APIs

tt2468:
I've created a simple async Python library for people who just want the simple stuff. Figured I would add it here.


Palakis:
@tt2468 Thanks!

PR #474, opened 2020-04-22
Suppress websocketpp logging entries

tt2468:
Regardless of whether debug is enabled or disabled, websocketpp seems to dump its log messages into the OBS console. These are annoying and clutter up logs. They are also redundant if debug logging is enabled.


tt2468:
It appears that currently the logging is shown on Linux but not on Windows.

PR #476, opened 2020-04-27
[Bug Fix] Fix FreeType2 source type recognition

tt2468:
OBS appears to be using a new format for the FreeType2 source, so we add the new kinds.

One thing I'm unsure on is if anything still uses the old naming stuff `text_ft2`/`text_ft2_v2`, since if not then we should not check.


Palakis:
@tt2468 
> One thing I'm unsure on is if anything still uses the old naming stuff text_ft2/text_ft2_v2, since if not then we should not check.

obs-websocket 4.8 targets OBS v25 so I guess you can get rid of these

PR #495, opened 2020-05-11
Requests: Add useDecibel bool to GetVolume and SetVolume

### LINKED ISSUE #441: GetVolume gives wrong values when slider under -35db
##### Issue type
Bug report (obs-websocket 4.7.0 with OBS Studio 25.1 - 64 bit)

##### Description
GetVolume returns values that do not correlate with neither the db or the the slider's position. 
For example :
0db = 1
-1db = 0.89
-2db = 0.79
-3db = 
...
-10db=0.31
-20db=0.099
-30db=0.031
-35db=0.0178
-40db=0.01
.... <--- Everything under this is under 0.01
-60db=0.0009
-90db=3.1708583264844492E-05

Seems to be a regression from previous versions.

Call GetVolume() (tested through obs-websocket-dotnet)

##### Technical information
- **Operating System** : Windows 10 1909 64 bit
- **OBS Studio version** : 25.1 64 bit

tt2468:
Gives a better, built-in option for people to use decibels of attenuation instead of amplitude. Default response/request format is amplitude.

Resolves #441

### DIFF
--- a/src/WSRequestHandler_Sources.cpp
+++ b/src/WSRequestHandler_Sources.cpp
@@ -156,12 +156,13 @@ RpcResponse WSRequestHandler::GetSourceTypesList(const RpcRequest& request)
 }
 
 /**
-* Get the volume of the specified source.
+* Get the volume of the specified source. Default response uses mul format, NOT SLIDER PERCENTAGE.
 *
 * @param {String} `source` Source name.
+* @param {boolean (optional)} `useDecibel` Output volume in decibels of attenuation instead of amplitude/mul.
 *
 * @return {String} `name` Source name.
-* @return {double} `volume` Volume of the source. Between `0.0` and `1.0`.
+* @return {double} `volume` Volume of the source. Between `0.0` and `1.0` if using mul, under `0.0` if using dB (since it is attenuating).
 * @return {boolean} `muted` Indicates whether the source is muted.
 *
 * @api requests
@@ -173,6 +174,11 @@ RpcResponse WSRequestHandler::GetVolume(const RpcRequest& request)
 {
 	if (!request.hasField("source")) {
 		return request.failed("missing request parameters");
+	}
+
+	bool useDecibel = false;
+	if (request.hasField("useDecibel")) {
+		useDecibel = obs_data_get_bool(request.parameters(), "useDecibel");
 	}
 
 	QString sourceName = obs_data_get_string(request.parameters(), "source");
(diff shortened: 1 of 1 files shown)


src/WSRequestHandler_Sources.cpp
```diff
@@ -187,33 +193,45 @@ RpcResponse WSRequestHandler::GetVolume(const RpcRequest& request)
...
+	if (request.hasField("useDecibel")) {
+		useDecibel = obs_data_get_bool(request.parameters(), "useDecibel");
+	}
+
 	QString sourceName = obs_data_get_string(request.parameters(), "source");
 	float sourceVolume = obs_data_get_double(request.parameters(), "volume");
 
-	if (sourceName.isEmpty() || sourceVolume < 0.0 || sourceVolume > 1.0) {
+	if ((useDecibel && sourceVolume > 0.0) ||
+(!useDecibel && (sourceVolume < 0.0 || sourceVolume > 1.0)) || (sourceName.isEmpty())) {
```
Palakis, line 234: - The "empty source name" check should go first
- You can break this complex `if` statement into several parts like this:

```cpp
const isNotValidDecibel = (useDecibel && sourceVolume > 0.0);
const isNotValidMul = (!useDecibel && (sourceVolume < 0.0 || sourceVolume > 1.0));
if (sourceName.isEmpty() || isNotValidDecibel || isNotValidMul) {
    // ...
}
```

src/WSRequestHandler_Sources.cpp
```diff
@@ -187,33 +193,45 @@ RpcResponse WSRequestHandler::GetVolume(const RpcRequest& request)
...
+*/
 RpcResponse WSRequestHandler::SetVolume(const RpcRequest& request)
  {
 	if (!request.hasField("source") || !request.hasField("volume")) {
 		return request.failed("missing request parameters");
 	}
 
+	bool useDecibel = false;
+	if (request.hasField("useDecibel")) {
+		useDecibel = obs_data_get_bool(request.parameters(), "useDecibel");
```
Palakis, line 227: The return value of `obs_data_get_bool` defaults to false if the value can't be found, so you can simplify this to:

```cpp
bool useDecibel = obs_data_get_bool(request.parameters(), "useDecibel");
```

without the `if` statement

src/WSRequestHandler_Sources.cpp
```diff
@@ -222,7 +240,12 @@ RpcResponse WSRequestHandler::SetVolume(const RpcRequest& request)
 		return request.failed("specified source doesn't exist");
 	}
 
-	obs_source_set_volume(source, sourceVolume);
+	if (!useDecibel) {
+		obs_source_set_volume(source, sourceVolume);
+	} else {
+		float mul = obs_db_to_mul(sourceVolume);
+		obs_source_set_volume(source, mul);
+	}
```
Palakis, line 248: You can simplify that whole `if` statement into this:

```cpp
if (useDecibel) {
    sourceVolume = obs_db_to_mul(sourceVolume);
}
obs_source_set_volume(source, sourceVolume);
```

PR #503, opened 2020-05-14
Requests: Add GetTransitionPosition

tt2468:
Gets the current position of the active transition. Works on manual and also auto transition modes. Originally added to #439 but removed since #439 requires OBS changes and wont be ready for 4.8.

### DIFF
--- a/src/WSRequestHandler_Transitions.cpp
+++ b/src/WSRequestHandler_Transitions.cpp
@@ -120,3 +120,22 @@ RpcResponse WSRequestHandler::GetTransitionDuration(const RpcRequest& request) {
 	obs_data_set_int(response, "transition-duration", obs_frontend_get_transition_duration());
 	return request.success(response);
 }
+
+/**
+ * Get the position of the current transition.
+ *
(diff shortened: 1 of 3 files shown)


src/WSRequestHandler_Transitions.cpp
```diff
@@ -120,3 +120,22 @@ RpcResponse WSRequestHandler::GetTransitionDuration(const RpcRequest& request) {
...
+ * Get the position of the current transition.
+ *
+ * @return {double} `position` current transition position. This value will be between 0.0 and 1.0. Note: Transition returns 1.0 when not active.
+ *
+ * @api requests
+ * @name GetTransitionPosition
+ * @category transitions
+ * @since 4.8.0
+ */
+RpcResponse WSRequestHandler::GetTransitionPosition(const RpcRequest& request) {
```
Palakis, line 134: You forgot to add this method to the headers of `WSRequestHandler` and in the request handlers map

tt2468: Oops. Its been a long day. Fixed.

PR #341, opened 2019-06-06
Added component management features

ilyes64:
*AddNewSource feature
*AddNewScene feature
*AddActiveChild feature

_Edit by Palakis:_
Closes #179

PR #524, opened 2020-05-26
Create scene base functionality

### LINKED ISSUE #522: Crating scenes and sources
##### Issue type
Feature request

##### Description
Hey there, currently working on a small scheduler/auto-pilot for obs, and was planning on using this solution for remote control, but was surprised by there not being functions to create scenes and sources.

The option is clearly there in ` libobs/obs-source.h` and `obs.h` as i could tell from taking a glance at OBS docs.

I noticed that in this project you only import frontend and module api. Is there a reason behind that, or am i missing something?

Also if that's a thing worth doing from your standpoint, i'd be happy to investigate and try write this.

ProstoSanja:
First command of propositions from #522 

Discussion: should there also be bool `activate-immediately` in request, to activate new scene after creation?

Testing: UI updates properly.

edit by tt: Closes #522

### DIFF
--- a/src/WSRequestHandler.cpp
+++ b/src/WSRequestHandler.cpp
@@ -46,6 +46,7 @@ const QHash<QString, RpcMethodHandler> WSRequestHandler::messageMap {
 	{ "SetCurrentScene", &WSRequestHandler::SetCurrentScene },
 	{ "GetCurrentScene", &WSRequestHandler::GetCurrentScene },
 	{ "GetSceneList", &WSRequestHandler::GetSceneList },
+	{ "CreateScene", &WSRequestHandler::CreateScene },
 	{ "SetSceneTransitionOverride", &WSRequestHandler::SetSceneTransitionOverride },
 	{ "RemoveSceneTransitionOverride", &WSRequestHandler::RemoveSceneTransitionOverride },
 	{ "GetSceneTransitionOverride", &WSRequestHandler::GetSceneTransitionOverride },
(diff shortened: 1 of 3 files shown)


tt2468:
Is this not similar functionality to #341 ?

Also if you can, please read over the upcoming new contrib guidelines https://github.com/Palakis/obs-websocket/blob/a1ea84b28669f679a2e914151bea09129ea67384/CONTRIBUTING.md

ProstoSanja:
Thanks for guidelines link, updated everything according to its rules.

Well both yeas and know. It does achieve the same result as part of 341, but this implementation is slightly cleaner and uses more correct naming scheme in request params.

Honestly i'm here for the functionality, and im making this pr, due to seeing no action happening with 341. As long as it or this gets merged, i wont mind.

tt2468:
Yep all good. All you need to do is change `scene-name` to `sceneName` and it looks good.

The commit title could also be a bit more descriptive IMO. like `Requests: Add CreateScene`

ProstoSanja:
Sorry for force pushes, just rebased onto master and applied your suggestions.

The file WSRequestHandler_Scenes.cpp actually seems a bit inconsistent in terms of cameCase/dashes because i see both used there. Is there plan to deal with that at some point(just pointing it out because it kinda had me confused)?

tt2468:
Yes its because some prs/features were created before we had standards on everything. The old OBSRemote protocol used kebab-case. obs-websocket 5.0 will be a breaking release which fixes all of the inconsistencies

ProstoSanja:
Cool, got it!

Also, fixed name in docs.

tt2468:
Oh also in terms of the optional bool to make the scene active, I say leave it out, since it can be done 
on the client by performing `SetCurrentScene` right after creating the scene. Eventually when batch requests are implemented, doing multiple requests will be easier.

PR #527, opened 2020-05-29
Requests: Sceneitem manipulation requests

tt2468:
Similar to #341 but starts from scratch, implementing all of the checks that need to be done.

Added requests:
- `AddSceneItem`
- `GetSceneItemList`

Closes #179

PR #535, opened 2020-06-07
Create source base functionality (modelled after #524)

togetherbeer:
hey. this adds a CreateSource request similar to #524 

CreateScene and CreateSource are equivalent right now, but it feels good to have both methods: to help with intuition / searchability for users of the library, and to allow for divergence / expansion down the line

also working on a PR adding CreateSceneItem({ parent, child }) so sources can be attached to scenes


tt2468:
The request type `CreateSource` (and its variants) is reserved for creating inputs (what the UI calls sources).

Also #527 already handles scene item manipulation. Unfortunately that means that this PR is a duplicate.

togetherbeer:
hey, cool. i saw that #341 had gotten held up a little and heard people asking about this feature

seemed like a cool thing to take a quick shot at, smaller PRs being easier to review and consider etc

nice to see #527. thanks!

PR #592, opened 2020-09-24
docs: 📝 list multiple available web clients

rodrigograca31:
### Description
list multiple available web clients instead of only t2t2

### Motivation and Context
because there's alternatives, that's a good thing and people should use which one they prefer.

### How Has This Been Tested?
yes

### Types of changes
documentation changes

### Checklist:


tt2468:
Looks good. I think there are a few more out there that could be added to the list, but we can let them make PRs on their own.

tt2468 (approved):

Issue #444, opened 2020-03-26
New request type to trigger scripts


Palakis:
##### Issue type

Feature Request

##### Description

Provide a way to trigger scripts from obs-websocket

karlthepagan:
Sending hotkey events into OBS would be sufficient for many plugins IMO.

tt2468:
As @karlthepagan noted, this functionality is able to be performed via #427 . As such, will close

Issue #427, opened 2020-03-08
HotKeys: requests and events


emendes-daitan:
##### Issue type
Feature request

##### Description
Provide the ability to register events and send requests related to Hotkeys. (the main goal is to send 'PressHotkey' request)
Possible contents:
- Events
  - Hotkeys
    - HotkeyPressed
    - HotkeyUpdated
- Requests
  - Hotkeys
    - GetHotkeys
    - SetHotkey
    - PressHotkey

tt2468:
The request functionality has been implemented in 5.0.0. Unfortunately, the event functionality is not possible without major OBS changes.

Issue #180, opened 2018-01-23
Feature request: Control OBS Scripts


lebaston100:
##### Issue type
Feature request

##### Description
It would be a nice additon, now with the release of obs 21 to work out some sort of way to control scripts(Start, Pass information).
I'm not into the obs source code so no idea how this can be done technically, but there is already a scripting function to register a hotkey: https://obsproject.com/docs/scripting.html#obs_hotkey_register_frontend

tt2468:
This can now be done via #595 's requests.

PR #595, opened 2020-09-30
Requests: Add hotkeys press events

LorenaGdL:
### Description

This PR adds two new requests that allow triggering hotkeys callback functions over websocket. The hotkey can be defined either by its unique name (the one used during registration) or by specifing the combination of keys associated to trigger it. In the latter, if the same combination has been set for multiple hotkeys, all of them will be triggered.

_Note_: the requests are named `ProcessHotkey `and not `PressHotkey` because no "physical" pressing event is sent/triggered, rather just the underlying callback functions are called/processed. Please let me know if you'd rather name the requests differently.

### Motivation and Context

This PR was motivated by the need to trigger script functions remotely. The requests can be easily used to trigger custom hotkeys registered within a script, and of course they also work with the default OBS hotkeys.

Related feature requests: #427, #444, #180

### How Has This Been Tested?

Tested OS(s): Windows 7 SP1, Windows 10 2004 (OBS 25.0.8)

The changes were tested by sending the requests via webpage using [obs-websocket-js](https://github.com/haganbmj/obs-websocket-js). For `ProcessHotkeyByName` request, both OBS default hotkeys and custom hotkeys (registered within scripts) were tested. For `ProcessHotkeyByCombination` request, tests were done with multiple key combinations (with and without modifiers), and associated to a single or multiple hotkey routines.

### Types of changes

### Checklist:

### DIFF
--- a/src/WSRequestHandler.cpp
+++ b/src/WSRequestHandler.cpp
@@ -44,6 +44,7 @@ const QHash<QString, RpcMethodHandler> WSRequestHandler::messageMap {
 	{ "BroadcastCustomMessage", &WSRequestHandler::BroadcastCustomMessage },
 
 	{ "ProcessHotkeyByName", &WSRequestHandler::ProcessHotkeyByName },
+	{ "ProcessHotkeyByCombination", &WSRequestHandler::ProcessHotkeyByCombination },
 
 	{ "SetCurrentScene", &WSRequestHandler::SetCurrentScene },
 	{ "GetCurrentScene", &WSRequestHandler::GetCurrentScene },
--- a/src/WSRequestHandler.cpp
+++ b/src/WSRequestHandler.cpp
@@ -43,8 +43,8 @@ const QHash<QString, RpcMethodHandler> WSRequestHandler::messageMap {
 
 	{ "BroadcastCustomMessage", &WSRequestHandler::BroadcastCustomMessage },
 
-	{ "ProcessHotkeyByName", &WSRequestHandler::ProcessHotkeyByName },
-	{ "ProcessHotkeyByCombination", &WSRequestHandler::ProcessHotkeyByCombination },
+	{ "TriggerHotkeyByName", &WSRequestHandler::TriggerHotkeyByName },
+	{ "TriggerHotkeyBySequence", &WSRequestHandler::TriggerHotkeyBySequence },
 
 	{ "SetCurrentScene", &WSRequestHandler::SetCurrentScene },
 	{ "GetCurrentScene", &WSRequestHandler::GetCurrentScene },
--- a/src/WSRequestHandler.cpp
+++ b/src/WSRequestHandler.cpp
@@ -43,6 +43,8 @@ const QHash<QString, RpcMethodHandler> WSRequestHandler::messageMap {
 
 	{ "BroadcastCustomMessage", &WSRequestHandler::BroadcastCustomMessage },
 
+	{ "ProcessHotkeyByName", &WSRequestHandler::ProcessHotkeyByName },
+
 	{ "SetCurrentScene", &WSRequestHandler::SetCurrentScene },
 	{ "GetCurrentScene", &WSRequestHandler::GetCurrentScene },
 	{ "GetSceneList", &WSRequestHandler::GetSceneList },
--- a/src/WSRequestHandler_General.cpp
+++ b/src/WSRequestHandler_General.cpp
@@ -345,3 +345,24 @@ RpcResponse WSRequestHandler::OpenProjector(const RpcRequest& request) {
 	obs_frontend_open_projector(type, monitor, geometry, name);
 	return request.success();
 }
+
+/**
+* Executes hotkey routine, identified by hotkey unique name
+*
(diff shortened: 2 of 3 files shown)


src/WSRequestHandler.cpp
```diff
@@ -43,6 +43,9 @@ const QHash<QString, RpcMethodHandler> WSRequestHandler::messageMap {
 
 	{ "BroadcastCustomMessage", &WSRequestHandler::BroadcastCustomMessage },
 
+	{ "ProcessHotkeyByName", &WSRequestHandler::ProcessHotkeyByName },
```
tt2468, line 46: I think `TriggerHotkeyByName` and `TriggerHotkeyBySequence` would be better names, as they fall more in line with how the OBS api references them.

src/WSRequestHandler_General.cpp
```diff
@@ -345,3 +345,69 @@ RpcResponse WSRequestHandler::OpenProjector(const RpcRequest& request) {
...
+	combo.modifiers = modifiers;
+	combo.key = obs_key_from_name(obs_data_get_string(request.parameters(), "key"));
+
+	if (!modifiers
+		&& (combo.key == OBS_KEY_NONE || combo.key >= OBS_KEY_LAST_VALUE)) {
+		return request.failed("Invalid key-modifier combination");
+	}
+
+	// Inject hotkey press-release sequence
+	obs_hotkey_inject_event(combo, false);
```
tt2468, line 408: Just curious, what's the purpose behind setting it `false` first?

src/WSRequestHandler_General.cpp
```diff
@@ -345,3 +345,69 @@ RpcResponse WSRequestHandler::OpenProjector(const RpcRequest& request) {
...
+	}
+	obs_hotkey_trigger_routed_callback(obs_hotkey_get_id(hk), true);
+	return request.success();
+}
+
+/**
+* Executes hotkey routine, identified by bound combination of keys. A single key combination might trigger multiple hotkey routines depending on user settings 
+*
+* @param {String} `key` Main key identifier (e.g. `OBS_KEY_A` for key "A"). Available identifiers [here](https://github.com/obsproject/obs-studio/blob/master/libobs/obs-hotkeys.h)
+* @param {Object (Optional)} `modifiers` Optional key modifiers object: `{"shift":true/false, "alt":true/false, "control":true/false, "command":true/false}`. False entries can be ommitted
```
tt2468, line 374: I would suggest the correct object document format, but it would be easier for me to just commit it. Mind if I do that?

src/WSRequestHandler_General.cpp
LorenaGdL: It's a good question and I expected it because it's definitely not intuitive. Explanation relates to how OBS `handle_binding()` function works, especially this [line](https://github.com/obsproject/obs-studio/blob/b9ea45e4982f4ec26b3f115f0e625289a8f26ae4/libobs/obs-hotkey.c#L1255). Apparently, when using modifiers in the combination, you need to do 2 passes of the function to properly set the` binding->modifiers_match` parameter (which is by default false, and is only potentially true after second pass). So that first call to the `inject_event()` helps initializing the param if needed.

To be honest, I had to dive deep into the OBS code and debug this request thoroughly because I wasn't expecting this line to be needed either, but it's the solution I found to make it work with modifiers.

src/WSRequestHandler.cpp
LorenaGdL: Not so sure about the "sequence" part. OBS refers to that as key combination. But anyway, I'm fine with whatever you prefer

src/WSRequestHandler_General.cpp
LorenaGdL: Sure, go ahead.
This is my first commit to the project and I tried to stick to your formats as much as I could, but bear with me and make all the changes you consider appropriate.

tt2468: Yeah all good, we don't ever expect people to be able to get it right the first time. obs-studio's PR guidelines are on another level of strict lol

src/WSRequestHandler.cpp
tt2468: I meant as in the `Trigger` language

src/WSRequestHandler_General.cpp
tt2468: Very interesting. Good work

LorenaGdL:
Thanks for the docs updates and the other fixes, I'll keep them in mind for future PRs

We need to adjust the TriggerHotkeyBySequence/TriggerHotkeyByCombination name, we ended up using different stuff in our PRs. Feel free to update whatever you want

tt2468 (approved):

LorenaGdL:
@tt2468 I guess you didn't see my message before merging. New docs refer to TriggerHotkeyByCombination but the request is actually called TriggerHotkeyBySequence. You need to change either of those, whichever you prefer.

Sorry for all the mess :S

tt2468:
I did read your message, but didn't understand it correctly. Oops

PR #609, opened 2020-11-03
CI: Remove BUILD_CAPTIONS flag

tt2468:
OBS has removed the `BUILD_CAPTIONS` flag from their repo. No reason to keep it


tt2468:
Closing for now because I realized we use the release versions of OBS, and so this is going to be broken until the next OBS release.

Palakis (approved):

Issue #624, opened 2020-12-04
SetSceneItemRender should handle duplicated sources in the same scene 


gdhgdhgdh:
##### Issue type
Feature request

##### Description
I have scenes where the `dj1` VLC media source exists in its original form and also as a `Copy -> Paste (Reference)` at a smaller size in the same scene so the video is perfectly synced, and more efficient than decoding twice.

This means that `dj1` appears twice in the list of Sources for the scene. If I rename either, the other is automatically renamed to the same thing.

When I call `SetSceneItemRender` with `source`: "dj1", `render`: false, only one of the two instances of `dj1` is hidden. I would happily pass the OBS `itemId` instead of the string `source` but `SetSceneItemRender` only supports string input :(

##### Technical information
- **Operating System** : Ubuntu 20.04
- **OBS Studio version** : 26.0.2 from Ubuntu PPA
- **obs-websocket version** : 4.8.0 .deb

tt2468:
Closing since this should have been closed via #628

PR #628, opened 2020-12-08
Request: add ID support for SetSceneItemRender

gdhgdhgdh:
This helps to address #624 and solves my particular use-case with minimal new `obs-websocket` code

### Description
`item` (an int) can now be passed instead of `source` (string) to `SetSceneItemRender` addressing the situation where the same source has been duplicated in a scene, and hence has the same name.

### How Has This Been Tested?
Verified on my Ubuntu 20.04 laptop - software was built locally using the scripts in the project `CI/` directory

### Checklist:

### DIFF
--- a/src/WSRequestHandler_SceneItems.cpp
+++ b/src/WSRequestHandler_SceneItems.cpp
@@ -393,7 +393,8 @@ RpcResponse WSRequestHandler::ResetSceneItem(const RpcRequest& request) {
 * Show or hide a specified source item in a specified scene.
 *
 * @param {String (optional)} `scene-name` Name of the scene the scene item belongs to. Defaults to the currently active scene.
-* @param {String} `source` Scene Item name.
+* @param {String (optional)} `source` Scene Item name.
+* @param {int (optional)} `item` Scene Item id
 * @param {boolean} `render` true = shown ; false = hidden
 *
 * @api requests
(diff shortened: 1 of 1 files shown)


src/WSRequestHandler_SceneItems.cpp
```diff
@@ -421,12 +424,21 @@ RpcResponse WSRequestHandler::SetSceneItemRender(const RpcRequest& request) {
...
 
-	OBSSceneItemAutoRelease sceneItem =
-		Utils::GetSceneItemFromName(scene, itemName);
-	if (!sceneItem) {
-		return request.failed("specified scene item doesn't exist");
-	}
+	OBSSceneItemAutoRelease sceneItem;
 
+	if (strlen(itemName)) {
+		OBSDataItemAutoRelease itemField = obs_data_item_byname(request.parameters(), "source");
```
Palakis, line 430: What's the use for this variable? I don't see it used anywhere in the scope block.

src/WSRequestHandler_SceneItems.cpp
```diff
@@ -421,12 +424,21 @@ RpcResponse WSRequestHandler::SetSceneItemRender(const RpcRequest& request) {
...
+		sceneItem = Utils::GetSceneItemFromName(scene, itemName);
+		if (!sceneItem) {
+			return request.failed("specified scene item name doesn't exist");
+		}
+	} else {
+		OBSDataItemAutoRelease itemField = obs_data_item_byname(request.parameters(), "item");
```
Palakis, line 436: Same question as above.

src/WSRequestHandler_SceneItems.cpp
```diff
@@ -402,16 +403,18 @@ RpcResponse WSRequestHandler::ResetSceneItem(const RpcRequest& request) {
 * @since 0.3
 */
 RpcResponse WSRequestHandler::SetSceneItemRender(const RpcRequest& request) {
-	if (!request.hasField("source") ||
-		!request.hasField("render"))
+	if (!request.hasField("render") ||
+		(!request.hasField("source") && !request.hasField("item"))
```
Palakis, line 407: It would be cleaner to have everything on one line. To make it more readable, you can write it like this:

```cpp
bool hasSourceOrItemParameter = ( request.hasField("source") || request.hasField("item") );
if (!request.hasField("render") || !hasSourceOrItemParameter) {
```

Palakis (changes requested):

gdhgdhgdh:
Aggh I completely forgot to follow up on this - your comments were :100: - that variable was dumb copy / paste on my part...

Palakis (approved):

PR #629, opened 2020-12-09
Requests: SetVolume: allow volumes > 1.0

marcan:
### Description

Allow volumes > 1.0 / > 0dB, which is legal in OBS (you can do this in Advanced Audio Properties). OBS allows up to +26dB gain, so we do the same. This corresponds to approximately 20x linear gain (actually 19.9526231497, but 20 is easier to explain and OBS doesn't care).

### Motivation and Context

This is required to be able to add audio gain to sources that have too low an audio level.

### How Has This Been Tested?

A previous version of this patch was tested with OBS master a few months ago, and used in production. I've rebased this on the current version, which added the decibel mode. I think the change is obvious enough that it does not warrant further testing.

### Types of changes

### Checklist:


marcan:
FWIW this has now been tested in production too.

tt2468:
Just a note for the future, only the documentation in the source code needs to be updated, as the docs files are automatically generated.

tt2468 (approved):

PR #641, opened 2020-12-24
General: Add ScenesCollection object

VMinute:
Adds a ScenesCollection object in the protocol definition,
replacing the current Array<String> return  with
Array<ScenesCollection>, keeping it more coherent with
other requests that return objects in the same format.
This will help automated code generation from comment.json
that otherwise would require ad-hoc handling for that specific
request.

Signed-off-by: Valter Minute <valter.minute@gmail.com>

### Description

Added new typedef and changed @return of WSRequestHandler::ListSceneCollections to generate a new object type named ScenesCollection instead of using string.
The changes impacts only generated documentation and comments.json, since the request already returns an array of objects.

### Motivation and Context

This change will make automated code generation of APIs in other languages more straightforward since the description of the API will match its actual return type.
Obviously an array of objects with a single string field can be simplified to a string array, but this will require ad-hoc handling in the API code and can still be done manually on API side.

### How Has This Been Tested?

Rebuilt the documentation following instructions in the README.
Changes are limited to source code comments.

Tested OS(s): 

Ubuntu 18.04

### Types of changes

Bug fix (non-breaking change which fixes an issue) 
Documentation change (a change to documentation pages)

### Checklist:
-   [X] I have read the [**contributing** document](https://github.com/Palakis/obs-websocket/blob/4.x-current/CONTRIBUTING.md).
-   [X] My code is not on the master branch.
-   [X] The code has been tested.
-   [X] All commit messages are properly formatted and commits squashed where appropriate.
-   [X] I have included updates to all appropriate documentation.

### DIFF
--- a/src/WSRequestHandler_SceneCollections.cpp
+++ b/src/WSRequestHandler_SceneCollections.cpp
@@ -3,6 +3,11 @@
 #include "WSRequestHandler.h"
 
 /**
+* @typedef {Object} `ScenesCollection`
+* @property {String} `sc-name` Name of the scene collection
+*/
+
+/**
  * Change the active scene collection.
  *
  * @param {String} `sc-name` Name of the desired scene collection.
@@ -59,8 +64,7 @@ RpcResponse WSRequestHandler::GetCurrentSceneCollection(const RpcRequest& reques
 /**
  * List available scene collections
  *
- * @return {Array<String>} `scene-collections` Scene collections list
- * @return {String} `scene-collections.*.sc-name` Scene collection name
+ * @return {Array<ScenesCollection>} `scene-collections` Scene collections list
  *
  * @api requests
  * @name ListSceneCollections
--- a/docs/generated/comments.json
+++ b/docs/generated/comments.json
@@ -476,6 +476,31 @@
     },
     {
       "subheads": [],
+      "typedef": "{Object} `ScenesCollection`",
+      "property": "{String} `sc-name` Name of the scene collection",
+      "properties": [
+        {
+          "type": "String",
+          "name": "sc-name",
+          "description": "Name of the scene collection"
+        }
+      ],
+      "typedefs": [
+        {
+          "type": "Object",
+          "name": "ScenesCollection",
+          "description": ""
+        }
+      ],
+      "name": "",
+      "heading": {
+        "level": 2,
+        "text": ""
+      },
+      "examples": []
+    },
+    {
+      "subheads": [],
       "typedef": "{Object} `Scene`",
       "property": [
         "{String} `name` Name of the currently active scene.",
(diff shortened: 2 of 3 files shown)


src/WSRequestHandler_SceneCollections.cpp
```diff
@@ -12,26 +17,32 @@
  * @category scene collections
  * @since 4.0.0
  */
-RpcResponse WSRequestHandler::SetCurrentSceneCollection(const RpcRequest& request) {
-	if (!request.hasField("sc-name")) {
+RpcResponse WSRequestHandler::SetCurrentSceneCollection(const RpcRequest &request)
+{
+	if (!request.hasField("sc-name"))
+	{
```
Palakis, line 23: The opening bracket of `if` blocks must be on the same line as the statement.

src/WSRequestHandler_SceneCollections.cpp
```diff
@@ -12,26 +17,32 @@
  * @category scene collections
  * @since 4.0.0
  */
-RpcResponse WSRequestHandler::SetCurrentSceneCollection(const RpcRequest& request) {
-	if (!request.hasField("sc-name")) {
+RpcResponse WSRequestHandler::SetCurrentSceneCollection(const RpcRequest &request)
```
Palakis, line 20: References and pointers must have their qualifier aligned with the type:
```suggestion
RpcResponse WSRequestHandler::SetCurrentSceneCollection(const RpcRequest& request)
```

src/WSRequestHandler_SceneCollections.cpp
```diff
@@ -12,26 +17,32 @@
...
+{
+	if (!request.hasField("sc-name"))
+	{
 		return request.failed("missing request parameters");
 	}
 
-	const char* sceneCollection = obs_data_get_string(request.parameters(), "sc-name");
-	if (!sceneCollection) {
+	const char *sceneCollection = obs_data_get_string(request.parameters(), "sc-name");
```
Palakis, line 27: Same as with references:
```suggestion
	const char* sceneCollection = obs_data_get_string(request.parameters(), "sc-name");
```

src/WSRequestHandler_SceneCollections.cpp
```diff
@@ -12,26 +17,32 @@
...
 	}
 
-	obs_queue_task(OBS_TASK_UI, [](void* param) {
-		obs_frontend_set_current_scene_collection(reinterpret_cast<const char*>(param));
-	}, (void*)sceneCollection, true);
+	obs_queue_task(
+		OBS_TASK_UI, [](void *param) {
+			obs_frontend_set_current_scene_collection(reinterpret_cast<const char *>(param));
+		},
+		(void *)sceneCollection, true);
```
Palakis, line 45: Why did it end up like this?

Palakis (changes requested):
Thanks for your contribution!
Before I can merge this, there are a few issues to address. This PR has too many unrelated changes and style issues.

src/WSRequestHandler_SceneCollections.cpp
VMinute: Sorry, I left my auto-formatter enabled and it was configured for a different coding style. I reverted all the code "changes" made by it and left only the relevant one that is in the comment at the beginning of the file.
Excuse me for the waste of time due to my distraction.

src/WSRequestHandler_SceneCollections.cpp
VMinute: Auto-formatting, my fault. Actually there are no relevant changes there, so I reverted that from my last commit.

VMinute:
Sorry, I didn't review the PR accurately and missed the style changes applied by auto-format in my editor. I removed all those non-relevant changes, leaving only the significant one (comment at the beginning).

src/WSRequestHandler_SceneCollections.cpp
```diff
@@ -2,6 +2,11 @@
 
 #include "WSRequestHandler.h"
 
+/**
+* @typedef {Object} `ScenesCollection`
```
Palakis, line 6: Now you may want to use this in the spec of `ListSceneCollections`

VMinute: Sounds like a good idea... :)
(sorry again for this messy thing, I am not used to github)

Palakis (approved):

PR #649, opened 2021-01-07
readme: Add Homebrew installation instructions

dnaka91:
Give a short hint that the project can be installed with Homebrew on
MacOS as alternative to installing manually.

### Description

I added a Homebrew (Cask) Formula a while back that allows to simply install `obs-websocket` with `brew install obs-websocket` as alternative to manual installation. I thought it might be nice to mention that in the README.md.

### Motivation and Context

Totally optional, but may be nice for people to know there is a convenient way to install the plugin.

### How Has This Been Tested?

Tested OS(s): MacOS

Didn't really test, as this is just a change to the docs. But used a markdown previewer to make sure the output looks proper and tested the Homebrew Formula of course, as it wouldn't have been merged otherwise.

### Types of changes

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,14 @@ WebSockets API for OBS Studio.
 
 Binaries for Windows, MacOS, and Linux are available in the [Releases](https://github.com/Palakis/obs-websocket/releases) section.
 
+### Homebrew
+
+If you're using MacOS you can use Homebrew for installation as well:
+
+```sh
+brew install obs-websocket
+```
+
 ## Using obs-websocket
 
 Here is a list of available web clients: (compatible with tablets and other touch interfaces)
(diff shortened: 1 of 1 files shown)


tt2468:
Just to confirm, are you committed to doing the appropriate updating to the homebrew formula when necessary? If it stops working in the future without a fix, we would of course remove the reference.

dnaka91:
Yes, I wouldn't mind keeping the formula up to date. After all releases don't seem to happen that often.

Sometimes may take me a few days maybe to update but usually it's a quick thing. And should I ever not want to update it anymore for whatever reason, I'll let you know.

I think there is even a way to tell homebrew to auto-update by adding some settings but the docs on that are rather sparse... I'll give it a try though.

Palakis (approved):

PR #667, opened 2021-02-02
Events: Remove sourceSettings from SourceCreated event

### LINKED ISSUE #580: SourceCreated event not providing 'sourceSettings'
##### Issue type
 - Bug report 

##### Description
The `SourceCreated` event doesn't seem to ever populate the `sourceSettings` property, possibly because the event is raised as soon as the source configuration window appears. I've tested it with `Image`, `Window Capture`, and `Game Capture` sources. It looks like when OBS is started the `SourceCreated` event is raised for each source when everything is being loaded, and it does have the settings then.

* Open OBS and start the websocket server and enable debug logging, if needed.
* Add some sources to a scene.
* Look at the `SourceCreated` updates in the log.

##### Technical information
- **Operating System** : Windows 10 Pro 2004
- **OBS Studio version** : 25.0.8
- **obs-websocket version** :  4.8.0

[2020-08-15 19-58-34.txt](https://github.com/Palakis/obs-websocket/files/5079816/2020-08-15.19-58-34.txt)

tt2468:
### Description

Removes the `sourceSettings` item from the `SourceCreated` event. Decided not to use the deprecation process since the parameter is always `{}` and would never have any users.
Closes #580

### Motivation and Context

The parameter is useless, as a freshly created source always has default settings, which do not get passed when obs_data_get_json() is called. Users should defer to using `GetSourceDefaultSettings`.

### How Has This Been Tested?

Tested OS(s): Ubuntu 20.04

### Types of changes

- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:

### DIFF
--- a/src/WSEvents.cpp
+++ b/src/WSEvents.cpp
@@ -1080,7 +1080,6 @@ void WSEvents::OnTransitionVideoEnd(void* param, calldata_t* data) {
  * @return {String} `sourceName` Source name
  * @return {String} `sourceType` Source type. Can be "input", "scene", "transition" or "filter".
  * @return {String} `sourceKind` Source kind.
- * @return {Object} `sourceSettings` Source settings
  *
  * @api events
  * @name SourceCreated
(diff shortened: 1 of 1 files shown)


Palakis:
@tt2468
> the parameter is always {} and would never have any users.

Is it also the case for sources created by `CreateSource`?

tt2468:
I suppose technically it does, but it's the only way that `sourceSettings` would be anything but {}.

PR #690, opened 2021-03-02
Requests: Add SetAudioTracks and GetAudioTracks

BlakeB415:
### Description

Adds SetAudioTracks and GetAudioTracks which allows toggling the audio output of a source to any of the 6 channels.

### Motivation and Context
Wanted to be able to control which source outputs audio to a track dynamically.

### How Has This Been Tested?

Tested OS(s): 
Windows

### Types of changes

New request/event (non-breaking)

### Checklist:

### DIFF
--- a/src/WSRequestHandler_Sources.cpp
+++ b/src/WSRequestHandler_Sources.cpp
@@ -408,7 +408,7 @@ RpcResponse WSRequestHandler::GetAudioTracks(const RpcRequest& request)
 /**
 * Get the mute status of a specified source.
 *
-* @param {String} `sourceName` Source name.
+* @param {String} `source` Source name.
 *
 * @return {String} `name` Source name.
 * @return {boolean} `muted` Mute status of the source.
--- a/src/WSRequestHandler_Sources.cpp
+++ b/src/WSRequestHandler_Sources.cpp
@@ -316,6 +316,97 @@ RpcResponse WSRequestHandler::SetVolume(const RpcRequest& request)
 }
 
 /**
+* Changes whether an audio track is active for a source.
+*
+* @param {String} `source` Source name.
+* @param {int} `track` Audio tracks 1-6.
+* @param {boolean} `active` Whether audio track is active or not.
+*
+* @api requests
+* @name SetTracks
+* @category sources
+* @since 4.0.0
+*/
+RpcResponse WSRequestHandler::SetAudioTracks(const RpcRequest& request)
+ {
+	if (!request.hasField("source") || !request.hasField("track") || !request.hasField("active")) {
+		return request.failed("missing request parameters");
+	}
+
+	QString sourceName = obs_data_get_string(request.parameters(), "source");
+	bool active = obs_data_get_bool(request.parameters(), "active");
+	int track = obs_data_get_int(request.parameters(), "track")-1;
+
+	if (sourceName.isEmpty()) {
+		return request.failed("invalid request parameters");
+	}
+
+	OBSSourceAutoRelease source = obs_get_source_by_name(sourceName.toUtf8());
+	if (!source) {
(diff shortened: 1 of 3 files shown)


src/WSRequestHandler_Sources.cpp
```diff
@@ -315,6 +315,97 @@ RpcResponse WSRequestHandler::SetVolume(const RpcRequest& request)
...
+* Changes whether an audio track is active for a source.
+*
+* @param {String} `source` Source name.
+* @param {int} `track` Audio tracks 1-6.
+* @param {boolean} `active` Whether audio track is active or not.
+*
+* @api requests
+* @name SetTracks
+* @category sources
+* @since 4.0.0
```
tt2468, line 328: We use `@since unreleased` for new requests that have not been put into a release yet.

src/WSRequestHandler_Sources.cpp
```diff
@@ -315,6 +315,97 @@ RpcResponse WSRequestHandler::SetVolume(const RpcRequest& request)
 	return request.success();
 }
 
+/**
+* Changes whether an audio track is active for a source.
+*
+* @param {String} `source` Source name.
```
tt2468, line 321: It'd be better to use the field `sourceName` instead of `source` to be more consistent with our current naming system.

src/WSRequestHandler_Sources.cpp
```diff
@@ -315,6 +315,97 @@ RpcResponse WSRequestHandler::SetVolume(const RpcRequest& request)
...
+ {
+	if (!request.hasField("source") || !request.hasField("track") || !request.hasField("active")) {
+		return request.failed("missing request parameters");
+	}
+
+	QString sourceName = obs_data_get_string(request.parameters(), "source");
+	bool active = obs_data_get_bool(request.parameters(), "active");
+	int track = obs_data_get_int(request.parameters(), "track")-1;
+
+	if (sourceName.isEmpty()) {
```
tt2468, line 340: Don't have to remove it if you don't want to, but since an empty source name would lead to failure to find a valid source anyway, it's not very necessary.

src/WSRequestHandler_Sources.cpp
```diff
@@ -315,6 +315,97 @@ RpcResponse WSRequestHandler::SetVolume(const RpcRequest& request)
...
+		return request.failed("invalid request parameters");
+	}
+
+	OBSSourceAutoRelease source = obs_get_source_by_name(sourceName.toUtf8());
+	if (!source) {
+		return request.failed("specified source doesn't exist");
+	}
+
+	uint32_t mixers = obs_source_get_audio_mixers(source);
+	uint32_t new_mixers = mixers;
```
tt2468, line 350: There shouldn't be any need to copy this variable

src/WSRequestHandler_Sources.cpp
```diff
@@ -315,6 +315,97 @@ RpcResponse WSRequestHandler::SetVolume(const RpcRequest& request)
...
+
+	if (active)
```
tt2468, line 352: Could the audio mixers be pushed out of range by calling the same request twice?

src/WSRequestHandler_Sources.cpp
```diff
@@ -315,6 +315,96 @@ RpcResponse WSRequestHandler::SetVolume(const RpcRequest& request)
...
+	obs_source_set_audio_mixers(source, mixers);
+
+	return request.success();
+}
+
+
+/**
+* Gets whether an audio track is active for a source.
+*
+* @param {String} `source` Source name.
```
tt2468, line 365: here too

BlakeB415: Oh, sorry. Missed that. My bad.

tt2468 (approved):

PR #692, opened 2021-03-04
Events: Fix multiple shutdown crashes

tt2468:
### Description

- Join with the WSServer ASIO thread to allow connections to be closed properly.
- Fixes crash caused by events failing to broadcast during OBS shutdown.

### Motivation and Context

Previously, the thread created to run the ASIO event loop would be created then ignored for the duration of the server's lifespan. However, OBS can sometimes close fast enough where the IO thread does not have enough time to close all sessions before it is killed. We join with the thread to force OBS to wait until all queued ASIO tasks have been completed.

After the server stops listening, it is unsafe to try to send messages to it. This now checks to make sure that the server is listening before trying to broadcast an event.

### How Has This Been Tested?

Tested OS(s): Ubuntu 20.04

### Types of changes

### Checklist:


Palakis (approved):
LGTM

PR #694, opened 2021-03-05
Requests: Add GetSourceActive

ruggi99:
### Description

Add GetSourceActive to know if a source is being showed in the final fix.
(Copy-pasted from GetAudioActive)

### Motivation and Context

In my context I want to know if a source is shown in the final mix.
This because I use the same sources (microphones) in multiple scenes and I want to know if my voice is audible while streaming / recording.
In my program I have three state for microphones. Active / Muted / Disabled (aka forced muted).
The only method to know if a mic is forced mute for now is to enumerate SceneItems of the current scene (eventually recursively). The method works great but I can't retrieve sources within a group. Other than that, I think it's a more appropriate way to use GetSourceActive.

### How Has This Been Tested?

Tested OS(s): 
Builded with Azure and tested quickly on my Windows 10

### Types of changes

### Checklist:

### DIFF
--- a/src/WSRequestHandler.cpp
+++ b/src/WSRequestHandler.cpp
@@ -124,6 +124,7 @@ const QHash<QString, RpcMethodHandler> WSRequestHandler::messageMap{
 	{ "GetMute", &WSRequestHandler::GetMute },
 	{ "SetMute", &WSRequestHandler::SetMute },
 	{ "ToggleMute", &WSRequestHandler::ToggleMute },
+	{ "GetSourceActive", &WSRequestHandler::GetSourceActive },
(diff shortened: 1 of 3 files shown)


ruggi99:
Tested too quickly I think.
Every source I tried is always active, even if it should not.
According to this: [https://obsproject.com/docs/reference-sources.html#c.obs_source_active](url) it should work as intended.
Can someone help me?

ruggi99:
I didn't think that during a transition all sources involved are active. So when I tried to get info on "SwitchScenes" event, the transition was still running... Trying on "TransitionEnd" fixed my error, so I think it's ready for review

tt2468 (approved):

PR #697, opened 2021-03-10
Requests: Add volumeDb field to SourceVolumeChanged Event

ruggi99:
### Description

Add volumeDb field to SourceVolumeChanged Event.

### Motivation and Context

Personally I work always with Decibel and when I get a `SourceVolumeChanged` Event, I always have to call `getVolume` (with useDecibel true) in order to get che correct value. This helps me a lot

### How Has This Been Tested?

Tested OS(s): Non tested yet. I except to test it when Azure finishes

### Types of changes

### Checklist:

-   [ ] The code has been tested.

### DIFF
--- a/src/WSEvents.cpp
+++ b/src/WSEvents.cpp
@@ -1145,6 +1145,7 @@ void WSEvents::OnSourceDestroy(void* param, calldata_t* data) {
  *
  * @return {String} `sourceName` Source name
  * @return {float} `volume` Source volume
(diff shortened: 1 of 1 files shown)


tt2468:
Ah right I forgot about that. It's because there is no real way to represent -inf in json, so it will just leave out the value. Need to add a check for `-inf` to set the decibel value to `-100`

ruggi99:
Thanks, I'll try

tt2468 (approved):

PR #706, opened 2021-03-21
docs: Mark from-scene as optional

dnaka91:
### Description

The `from-scene` property of the `TransitionBegin` and `TransitionVideoEnd` is optional but is not marked as such in the docs. 

Originally discovered in [obws](https://github.com/dnaka91/obws/pull/8).

### Motivation and Context

As this field wasn't marked as optional I added it as required field in `obws` which causes deserialization to fail if it's missing. It's already fixed there but should be updated here for correctness and future library writers.

### How Has This Been Tested?

Tested OS(s): Tested through the `obws` library with latest OBS and obs-websocket.

### Types of changes

### Checklist:


tt2468:
Your PR seems to include quite a few unnecessary changes: https://github.com/Palakis/obs-websocket/pull/706/files

The only file that needs changing is `WSEvents.cpp`. Everything else will be automatically committed once the PR is merged.

Also, is it truly optional or is it never sent? Because according to [here](https://github.com/Palakis/obs-websocket/blob/4.x-current/src/Utils.cpp#L501), it might just never be sent. In which case we should remove the field entirely.

dnaka91:
True, it's very unlikely that there will ever be a transition to the same scene right?

I ran the docs generation manually and included the changes. Will revert that part back then.
Should I revert the whitespace trimming changes as well?

tt2468:
> Should I revert the whitespace trimming changes as well?

Not strictly necessary, just the generated docs files and whatever `docs/npm-shrinkwrap.json` is

dnaka91:
Oh I'm so sorry. I didn't even realize that file slipped in there. Removed it now together with the unneeded docs changes 🙇

tt2468 (approved):

PR #707, opened 2021-03-21
requests: Add RemoveSource and RemoveScene

dnaka91:
### Description

Currently obs-websocket has requests to add new sources and scenes but no way of deleting them again. This adds the two new requests `RemoveSource` and `RemoveScene` to do so.

`RemoveScene` is almost identical to `RemoveSource`, just has an extra safe guard to check the source type before removing.

### Motivation and Context

Give more control to allow deleting things as well. Especially helpful in my `obws` library to automatically set up OBS for integration tests and remove all the test elements from OBS afterwards.

### How Has This Been Tested?

Tested OS(s): MacOS

Tested it by implementing the new requests in my `obws` library and executed them to check the results. Only tested with scenes so far and they could be removed without errors or crashes.

### Types of changes

### Checklist:


Palakis (commented):
Thank you for these additions! LGTM, although there are a few nitpicks in the code. I'm addressing these myself and showcasing them here for reference.

src/WSRequestHandler_Sources.cpp
```diff
@@ -76,17 +76,48 @@ RpcResponse WSRequestHandler::CreateSource(const RpcRequest& request)
...
+	if (sourceName.isEmpty()) {
+		return request.failed("invalid request parameters");
+	}
+
+	OBSSourceAutoRelease source = obs_get_source_by_name(sourceName.toUtf8());
+	if (!source) {
+		return request.failed("specified source doesn't exist");
+	}
+
+	obs_source_remove(source);
```
Palakis, line 116: This call works, but if the source is referenced in one or more scenes, remove the source directly doesn't trigger a UI update of the sources list in OBS.

For input sources, what do you think of checking first if it is used somewhere, and if so return an error message?

Palakis (changes requested):

src/WSRequestHandler_Sources.cpp
dnaka91: It would be nice to have a `force` flag then which would allow to remove it despite the side effects.

But as this would cause a weird state in the UI... Maybe better not to.

tt2468:
Closing this as it will not be able to be implemented via 4.9.1. We will be making the appropriate PRs to the obs-studio repository for 5.0.0.

PR #791, opened 2021-07-03
WSRequestHandler_Sources: fix comment name for Set/GetAudioTrack

dvangennip:
### Description
Functions were named SetTracks and GetTracks rather than SetAudioTracks and GetAudioTracks as in the actual code, so this mismatch gets picked up by the protocol docs (but not anymore).

### Motivation and Context
Docs were incorrectly listing GetTracks and SetTracks, requests that don't work.
Fixes #789 

### How Has This Been Tested?
Only comment changes, no further compiling or testing done.

### Types of changes
 - Documentation change (a change to documentation pages)

### Checklist:
-   [x] I have read the [**contributing** document](https://github.com/Palakis/obs-websocket/blob/4.x-current/CONTRIBUTING.md).
-   [x] My code is not on the master branch.
-   [ ] The code has been tested.
-   [x] All commit messages are properly formatted and commits squashed where appropriate.
-   [x] I have included updates to all appropriate documentation.

PR #794, opened 2021-07-12
Changed Set/GetTracks to Set/GetAudioTracks properly

AdelaFoofur82:
### How Has This Been Tested?

Tested OS(s): 

### Checklist:


tt2468:
Please follow the commit format for commit names.

Also, generated documents should not be included in commits. Please change the documentation via the source code: https://github.com/Palakis/obs-websocket/blob/4.x-current/src/WSRequestHandler_Sources.cpp#L318

dvangennip:
The change here only updates the protocol.md file, which gets generated from source file data. Initially, I did a similar thing in #791 so it looks like both patches address the same issue in different ways.

PR #842, opened 2021-10-24
Build: Make scripts executable and fix dep links

dnaka91:
### Description

Gives back the executable flag to the MacOS specific scripts. Also, adds the new dependencies `QtNetwork` and `QtSvg` to the dependency linking step so the binary doesn't target the `/tmp/obsdeps` folder.

### Motivation and Context

Required to make building on MacOS easy as the scripts seem to work just fine even after all the changes in `master` compared to `4.x-current`.

### How Has This Been Tested?

Tested OS(s): MacOS

Using the scripts already for several months to test my client library against the latest 5.0 version changes for MacOS.

### Types of changes

### Checklist:

-  [x] I have read the [**contributing** document](https://github.com/Palakis/obs-websocket/blob/4.x-current/CONTRIBUTING.md).
-  [x] My code is not on the master branch.


tt2468:
Merging for now, but I'm not sure if it will be used in CI. Might move to the `scripts` directory if it doesn't get used by CI

PR #874, opened 2021-12-17
docs: Fix `Array` type in markdown

dnaka91:
### Description

Fix the docs generator by replacing `<` with `&lt;` and `>` with `&gt;` for the **Type** column in request and response tables.

### Motivation and Context

The `<` and `>` pair has a special meaning in markdown, being interpreted as HTML tags. That caused the values like `Array<String>` to be displayed as just `Array`. Therefore, they needed to be replaced by their respective `&lt;` and `&gt;` codes.

It is now done for the **Type** column but may need to be extended into other fields that potentially have some form of `<Something>` content.

### How Has This Been Tested?

Tested OS(s): MacOS

Ran the `build_docs.sh` script and verified the expected outputs in `protocol.md`.

### Types of changes

### Checklist:

### DIFF
--- a/docs/docs/generate_md.py
+++ b/docs/docs/generate_md.py
@@ -178,15 +178,17 @@ def get_requests(requests):
             if request['requestFields']:
                 ret += requestFieldHeader
             for requestField in request['requestFields']:
+                valueType = requestField['valueType'].replace('<', "&lt;").replace('>', "&gt;")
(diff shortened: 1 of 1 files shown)


tt2468:
`valueType` is used here too: https://github.com/obsproject/obs-websocket/blob/25707958d18f872b7e7ae751165d9cd416245770/docs/docs/generate_md.py#L244

Would you be able to handle that too?

dnaka91:
Oh, good catch! Let's do for that one too 👍

PR #878, opened 2021-12-17
docs: Fix `scenes` type from `String` to `Object`

dnaka91:
### Description

Change the type of the `scenes` field of the `GetSceneList` request from `Array<String>` to ~~`Array<Scene>`~~ `Array<Object>`.

Where or how would you like me to document the fields of the `Scene` object?

### Motivation and Context

The request returns an array of scene objects but was documented as returning an array of strings instead.

### How Has This Been Tested?

Tested OS(s): 

### Types of changes

### Checklist:

### DIFF
--- a/src/requesthandler/RequestHandler_Scenes.cpp
+++ b/src/requesthandler/RequestHandler_Scenes.cpp
@@ -22,7 +22,7 @@ with this program. If not, see <https://www.gnu.org/licenses/>
 /**
  * Gets an array of all scenes in OBS.
  *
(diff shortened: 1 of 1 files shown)


tt2468:
The correct usage here is `Array<Object>`. The type fields are for json types only.

dnaka91:
Ah right, my bad. Still need to document the `Scene` type somewhere though, right?
So far, the only way to figure out the available fields is by diving into the source code.

PR #879, opened 2021-12-17
docs: Improve generated docs formatting

dnaka91:
### Description

Update the docs generation to make the docs markdown (and all other README.md as well actually) pass the default settings of markdownlint. With a few tiny exceptions here and there.

The **line length** rule is disabled globally as most editors do automatic line wrapping for us anyways.

### Motivation and Context

[markdowlint](https://github.com/DavidAnson/markdownlint) is a nice tool to check for common best practices that keep the Markdown clean and readable. Most of the code is generated and it probably _doesn't really matter_, but still it feels better to have a nice and clean output.

### How Has This Been Tested?

Tested OS(s): 

### Types of changes

### Checklist:


docs/docs/generate_md.py
```diff
@@ -160,7 +159,7 @@ def get_requests(requests):
         if not len(requestsOut):
             continue
         categoryFragment = get_fragment(category)
-        ret += '\n\n## {}\n\n'.format(category)
+        ret += '## {} requests\n\n'.format(category)
```
tt2468, line 162: Should `requests` be capitalized?

src/requesthandler/RequestHandler_Inputs.cpp
```diff
@@ -549,6 +549,7 @@ RequestResult RequestHandler::SetInputAudioSyncOffset(const Request& request)
  * Gets the audio monitor type of an input.
  *
  * The available audio monitor types are:
+ *
```
tt2468, line 552: Can you go over the rest of the documentation and make sure it's formatted correctly?

dnaka91: Sorry, will finally get back to this now.
Probably a few updates needed here and there but should be quick overall.

dnaka91:
Just pulled in the latest changes from master and adjusted remaining issues. All markdown files in the project look good now.

As a plus, I added a new workflow to the project that updates docs and runs `markdownlint-cli` on all Markdown files, so we directly see in the CI if some formatting is off.

Issue #868, opened 2021-12-06
Feature Request: API to get monitor information


wjzhou:
### Feature Request Type

RPC Request

### Requested Feature

Note: I know OBS doesn't expose this API, but the implementation inside OBS is less than 44 lines. (Even consider the window specific implementation, it is still a relatively easy function)
https://github.com/obsproject/obs-studio/blob/00e6f6e023549f6acfb5415c319edaeadb32945e/UI/window-basic-main.cpp#L4982-L5026

In short term could we reimplement this in obs-websocket? (The code only has dependency on QT, which we already depend on). long term, I will try to upstream the change to the obs plugin api.
I can create pull request if you think we will consider accepting a change like this.

---
API:
### GetMonitorList

Get information of monitors.

**Request Fields:**
No specified parameters.

**Response Items:**

| Name | Type  | Description |
| ---- | :---: | ------------|
| `monitors` | _Array<Object>_ | Array of monitors |
| `monitors.*.name` | _String_ | monitor name |
| `monitors.*.width` | _int_ | monitor width |
| `monitors.*.height` | _int_ | monitor height |
| `monitors.*.x` | _int_ | monitor x |
| `monitors.*.y` | _int_ | monitor y |

### Requested Feature Usage Scenario

Currently, I use `OpenProjector` to open projectors on multiply monitors. (With hard coded `monitor`)

However, Occasionally, Windows will reorder the monitors on reboot and the monitor number is wrong.

In the OBS GUI, I can use the monitor's name to differential the monitors, this request will help to expose that information to the websocket client.

PR #889, opened 2022-01-16
Requests: Add GetMonitorList

csunday95:
Adds a new request `GetMonitorList` that returns a json Array of
objects containing data about connected monitors. See Issue #868

### Description
Adds `GetMonitorList` endpoint as requested in #868 

### Motivation and Context
Adds the ability for clients to get monitor information. This can be useful for changing client behavior based on monitor resolution, position, presence of multiple monitors etc. 

### How Has This Been Tested?
 - tested end to end from a python script using simpleobsws (V1.3.1)
Tested OS(s): Linux, Ubuntu 21.10

### Types of changes

### Checklist:
-  [ x ] I have read the [Contributing Guidelines](https://github.com/obsproject/obs-websocket/wiki/Contributing-Guidelines).
-  [ x ] All commit messages are properly formatted and commits squashed where appropriate.
-  [ x ] My code is not on the `master` branch.
-  [ x ] The code has been tested.
-  [ x ] I have included updates to all appropriate documentation.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,6 +1,6 @@
 # obs-websocket documentation
 
-This is the documentation for obs-websocket. Run build_docs.sh to auto generate the latest docs from the `src` directory. There are 3 components to the docs generation:
+This is the documentation for obs-websocket. Run `build_docs.sh` to auto generate the latest docs from the `src` directory. There are 3 components to the docs generation:
 - `comments/comments.js`: Generates the `work/comments.json` file from the code comments in the src directory.
 - `docs/process_comments.py`: Processes `work/comments.json` to create `generated/protocol.json`, which is a machine-readable documentation format that can be used to create obs-websocket client libraries.
 - `docs/generate_md.py`: Processes `generated/protocol.json` to create `generated/protocol.md`, which is the actual human-readable documentation.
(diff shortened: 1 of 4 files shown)


csunday95:
points of discussion:
  - the OBS implementation mentioned in #868 does some sem-elaborate, platform specific stuff to get "more useful" human readable names, would replicating that behavior be desirable?
  - There doesn't seem to be a nice way to document the inner members of the return array objects with the current documentation, should I just add it to the description?

src/requesthandler/RequestHandler_General.cpp
```diff
@@ -92,6 +96,42 @@ RequestResult RequestHandler::GetStats(const Request&)
...
+/**
+ * Gets a list of connected monitors and information about them.
+ *
+ * @responseField monitors | Array<Object> | a list of detected monitors with some information
+ *
+ * @requestType GetMonitorList
+ * @complexity 2
+ * @rpcVersion -1
+ * @initialVersion 5.0.0
+ * @category general
```
tt2468, line 108: This would be better placed in the Ui category

tt2468 (changes requested):

src/requesthandler/RequestHandler_General.cpp
csunday95: moved to UI caegory

waiki510 (approved):

PR #923, opened 2022-03-11
Server: Implement 'Lock to IPv4' option from v4.x

ChrisTallon:
Also re-enables ability to connect using IPv6.

### Description

Copied and modified the relevant code from v4.x to reimplement the "Lock to IPv4" option in Server Settings.

Care was taken to preserve the same behaviour as 4.x, as was created by https://github.com/obsproject/obs-websocket/issues/531

The default value for 'Lock to IPv4' is true to ensure nothing changes after an upgrade to this code.

### Motivation and Context

Re-enables ability to connect using IPv6. This existed in v4.x but disappeared in the v5 rewrite.

### How Has This Been Tested?

Tested OS(s): 

Kubuntu 21.10
OBS Studio: 27.2.2-0obsproject1~impish

Ensured that the plugin behaved as the 4.x version by monitoring the TCP connection using the command "ss -plunt".

Example output with Lock to IPv4 selected:

tcp   LISTEN 0      4096                            0.0.0.0:4455       0.0.0.0:*    users:(("obs",pid=6059,fd=14))

And with it unselected:

tcp   LISTEN 0      4096                                  *:4455             *:*    users:(("obs",pid=6059,fd=14))

(The IPv6 bindings also listen on IPv4).

### Types of changes

### Checklist:

-  [x ] I have read the [Contributing Guidelines](https://github.com/obsproject/obs-websocket/wiki/Contributing-Guidelines).
-  [x ] All commit messages are properly formatted and commits squashed where appropriate.
-  [x ] My code is not on the `master` branch.
-  [x ] The code has been tested.
-  [x ] I have included updates to all appropriate documentation.


tt2468:
The original implementation was arguably just a patch for a very niche situation. Looking at the code though, I see that I explicitly bound the server to v4, which was a mistake. Did you actually require explicit ipv4 bindings? I think I might rather relegate this functionality to a command line switch.

ChrisTallon:
I don't require IPv4 bindings, my motivation was to re-enable IPv6 while keeping the existing behaviour. Here is a command line switch version that changes the current behaviour. Should I update this PR or create a new one?

https://github.com/ChrisTallon/obs-websocket/commit/133f66e7db2d46128b462e6d57dae4b6c9401862

tt2468:
We can just create a new PR for the change.

PR #966, opened 2022-07-24
fix(doc): auth string

GoooIce:
### Description

fix protocol doc auth string

### Motivation and Context

only doc, do not have code.

### How Has This Been Tested?

Tested OS(s): 
test with nodejs

### Checklist:

-  [ x] I have read the [Contributing Guidelines](https://github.com/obsproject/obs-websocket/wiki/Contributing-Guidelines).
-  [ x] All commit messages are properly formatted and commits squashed where appropriate.
-  [x ] My code is not on `master` or a `release/*` branch.
-  [x ] The code has been tested.
-  [ x] I have included updates to all appropriate documentation.

### DIFF
--- a/src/obs-websocket.h
+++ b/src/obs-websocket.h
@@ -26,7 +26,7 @@ with this program. If not, see <https://www.gnu.org/licenses/>
 #include "utils/Obs.h"
 #include "plugin-macros.generated.h"
 
-class Config;
+struct Config;
 typedef std::shared_ptr<Config> ConfigPtr;
 
 class EventHandler;


tt2468:
What is the purpose of changing this string in the docs? The one currently in there should be functional. Additionally, your commits are malformed and need to follow our commit guidelines. Finally, your string change should be done in <https://github.com/obsproject/obs-websocket/blob/master/docs/docs/partials/introduction.md> and not in the generated docs file, as it would just be overwritten.

PR #1097, opened 2023-02-02
Replace headers-only libraries submodules by prefix/system-wide install

tytan652:
### Description

Requires to work on obs-studio repo:
- https://github.com/obsproject/obs-deps/pull/155 (Asio 1.12.1)
- https://github.com/obsproject/obs-studio/pull/8194

Replace headers-only libraries submodules by prefix/system-wide install.

### Motivation and Context

Reduce the number of submodules in obs-studio repo and provides those libraries in obs-deps.

### How Has This Been Tested?

Tested OS(s): 

obs-studio CI on my fork (https://github.com/tytan652/obs-studio/pull/10) with obs-websocket modified to be aware of obs-deps and custom-made finders for Asio and WebSocket++.

### Types of changes

### Checklist:

-  [ ] The code has been tested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,11 +9,7 @@ if(NOT ENABLE_WEBSOCKET OR NOT ENABLE_UI)
 endif()
 
 # Submodule deps check
-if(NOT
-   (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/deps/json/CMakeLists.txt
-    AND EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/deps/websocketpp/CMakeLists.txt
-    AND EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/deps/qr/cpp/QrCode.hpp
-    AND EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/deps/asio/asio/include/asio.hpp))
+if(NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/deps/qr/cpp/QrCode.hpp)
   obs_status(FATAL_ERROR "obs-websocket submodule deps not available.")
 endif()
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -9,32 +9,24 @@ if(NOT ENABLE_WEBSOCKET OR NOT ENABLE_UI)
...
-    CACHE INTERNAL "")
-add_subdirectory(deps/json)
+# Find nlohmann JSON
+find_package(nlohmann_json 3 REQUIRED)
+
+# Find WebSocket++
+find_package(Websocketpp 0.8 REQUIRED)
+
+# Find Asio
+find_package(Asio 1.12.1 REQUIRED)
```
RytoEX, line 29: Should this use the `EXACT` option to ensure that we do not accidentally pick up a different version?

RytoEX (commented):
A question about find_package syntax.

CMakeLists.txt
tytan652: All Linux-related `find_package` should operate with the assumption that we can't enforce a exact version but only a minimum.

We control Windows, macOS and Flatpak deps version but not Ubuntu's (and other distros').

tt2468: As far as I know, the version limitations are mainly for MacOS and specifically for ASIO. I'm not aware of any intentional breaking changes in any recent history for any of these dependencies.

tytan652: I though we were talking about minimums on the three deps.

I placed them mostly to avoid building OBS with too old version of those libs mostly on old Linux distros.

tytan652: I can remove them, if you consider this is really unneeded.

tt2468: I think it's fair to keep them as is. I was mainly responding to whether or not we should use the `EXACT` keyword.

RytoEX: It's fine. My question has been answered. We control the version on the system where it's problematic, and using `EXACT` would probably be too burdensome for non-Flatpack Linux builds.

RytoEX (approved):
Looks good to me. Discussed off-thread that this is good to merge.

PR #1111, opened 2023-03-16
requesthandler: Fix existing input/scene detection in create

rjgtav:
### Description

Allow re-creating an input/scene with the same name in case the previous instance has already been marked for deletion.

### Motivation and Context

Deleting and re-creating an input or scene by using the same name, in consecutive requests, currently throws an "already exists", despite those items already being marked for deletion in OBS.

Here's the code snippet, using OBS Websocket's javascript bindings, for triggering the error:
```
import OBSWebSocket from 'obs-websocket-js';

const obs = new OBSWebSocket();
const sceneName = 'Scene';

await obs.call('RemoveScene', { sceneName });
await obs.call('CreateScene', { sceneName }); // Will throw "A source already exists by that scene name."
```

### How Has This Been Tested?

Tested OS(s): Debian 11

### Types of changes

### Checklist:

### DIFF
--- a/src/requesthandler/RequestHandler_Inputs.cpp
+++ b/src/requesthandler/RequestHandler_Inputs.cpp
@@ -149,7 +149,7 @@ RequestResult RequestHandler::CreateInput(const Request &request)
 
 	std::string inputName = request.RequestData["inputName"];
 	OBSSourceAutoRelease existingInput = obs_get_source_by_name(inputName.c_str());
-	if (existingInput)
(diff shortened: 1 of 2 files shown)


tt2468:
Unfortunately, this cannot be merged. The old behavior which would allow this to happen was undefined and not intended, and as of recently, we've changed the underlying libobs functions to deduplicate names automatically. As such, in both cases, this kind of change would produce behavior which is arguably unacceptable in an API like this.

PR #1134, opened 2023-05-07
RequestHandler: Fix memory leak when setting streaming service

norihiro:
### Description

Replaced `OBSService` type with `OBSServiceAutoRelease` in `RequestHandler::SetStreamServiceSettings`.

### Motivation and Context

Memory leaks when setting streaming service using obs-websocket.

### How Has This Been Tested?

Tested OS(s): Ubuntu 22.04

Steps to test the memory-leak is as below.

1. Start obs with clean configuration
   ```
   export XDG_CONFIG_HOME=$(mktemp -d /dev/shm/obs-XXXXXX)
   obs &
   ```
   1. In the wizard, configure `Optimize just for recording, I will not be streaming`.
2. Enable WebSocket server
   1. also disable password for simple testing.
4. Run these lines in python3
   ```
   import obsws_python
   cl = obsws_python.ReqClient(host='localhost', port=4455)
   cl.send('SetStreamServiceSettings', {
       'streamServiceType': 'rtmp_custom',
       'streamServiceSettings': {
           'server': 'rtmp://localhost/live',
           'key': 'secret',
           }})
   ```
5. Exit obs

### Types of changes

### Checklist:

### DIFF
--- a/src/requesthandler/RequestHandler_Config.cpp
+++ b/src/requesthandler/RequestHandler_Config.cpp
@@ -599,9 +599,9 @@ RequestResult RequestHandler::SetStreamServiceSettings(const Request &request)
 
 		obs_service_update(currentStreamService, newStreamServiceSettings);
 	} else {
-		// TODO: This leaks memory. I have no idea why.
-		OBSService newStreamService = obs_service_create(requestedStreamServiceType.c_str(), "obs_websocket_custom_service",
-								 requestedStreamServiceSettings, nullptr);
+		OBSServiceAutoRelease newStreamService = obs_service_create(requestedStreamServiceType.c_str(),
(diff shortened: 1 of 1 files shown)


tt2468:
Are you sure it leaks currently? I very explicitly chose OBSService as there was weird behavior with service reference counting that required it.

norihiro:
I've added steps to reproduce in the description of this PR.

Something I noticed weird is that `SetStreamServiceSettings` at first time will leak the memory but 2nd time and later won't.
This is because `SetStreamServiceSettings` will not re-create the object if the service-type matches.

If you have a doubt for the reference count, how about using `obs_service_t *` instead of `OBSService` and intentionally calls `obs_service_release`?

PR #1167, opened 2023-09-16
cmake: Update library and target names for qrcodegen

PatTheMav:
### Description
Changes identifiers for qrcodegen to use unified identifier provided by CMake packages.

### Motivation and Context
qrcodegen is identified as such (without the "lib" prefix) and as the target "qrcodegencpp::qrcodegencpp" by the CMake package generated by obs-deps.

Requires https://github.com/obsproject/obs-deps/pull/214 to be merged and released.
Needs to be integrated into https://github.com/obsproject/obs-studio/pull/9601.

### How Has This Been Tested?
Tested by building OBS Studio with websocket support and all related changes applied.

Tested OS(s): Windows 11, macOS 13.5.2, Ubuntu 22.10

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,9 @@ find_qt(COMPONENTS Core Widgets Svg Network)
 find_package(nlohmann_json 3 REQUIRED)
 
 # Find qrcodegencpp
-find_package(Libqrcodegencpp REQUIRED)
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
+find_package(qrcodegencpp REQUIRED)
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG OFF)
 
 # Find WebSocket++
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -18,7 +18,8 @@ find_qt(COMPONENTS Core Widgets Svg Network)
 find_package(nlohmann_json 3 REQUIRED)
 
 # Find qrcodegencpp
-find_package(Libqrcodegencpp REQUIRED)
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
```
RytoEX, line 21: Should this be toggled back OFF after the subsequent `find_package` call, or does it not matter?

PatTheMav: It doesn't matter _effectively_ but logically it should be reset, because I wouldn't say that we can make that same conscious decision for the other dependencies.

RytoEX (approved):
Looks good to me.

PR #1259, opened 2024-10-09
fix failing build

mildsunrise:
### Description
current code in master fails to build because `Json.cpp` uses `std::filesystem` without including `<filesystem>`

### Types of changes

### Checklist:
-  [X] I have read the [Contributing Guidelines](https://github.com/obsproject/obs-websocket/wiki/Contributing-Guidelines).
-  [X] All commit messages are properly formatted and commits squashed where appropriate.
-  [X] My code is not on `master` or a `release/*` branch.
-  [X] The code has been tested.

### DIFF
--- a/src/utils/Json.cpp
+++ b/src/utils/Json.cpp
@@ -18,6 +18,7 @@ with this program. If not, see <https://www.gnu.org/licenses/>
 */
 
 #include <fstream>
+#include <filestream>
 
 #include "Json.h"
 #include "plugin-macros.generated.h"


tt2468:
On what environment is build failing for you?

mildsunrise:
it's been half a year. I don't know the exact environment that this failure happened on, but this mistake seemed trivial enough that i didn't debug back then. If it helps I was using Arch Linux, and building obs-studio from latest master at that time. I'll try to reproduce...

tt2468:
Since this has been open for a while without anyone else reporting the same issue, I'll go ahead and close this for now. If the build error reoccurs for someone then we can consider reopening this.

PR #1265, opened 2024-10-23
Fix profile functions that direct call a method instead of the API

### LINKED ISSUE #1264: Bug: cannot create new profile in the beta for 31
### Operating System Info

Other

### Other OS

mac sequoia 15.0.1

### OBS Studio Version

Other

### OBS Studio Version (Other)

31.0.0-beta1

### obs-websocket Version

Git

### OBS Studio Log URL

https://obsproject.com/logs/4eZFCKKSMmx2D0lt

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

A new profile is created

### Current Behavior

No new profile is created.

Log file of:

> QMetaObject::invokeMethod: No such method OBSBasic::NewProfile(QString)

### Steps to Reproduce

1. Connect a websocket
2. Run `obs.call('CreateProfile', { profileName: 'dave' });`
3. Observe no new profile

### Anything else we should know?

Creating a new profile via OBS itself works fine.

The response via the Websocket is:

> {"d":{"requestId":"2","requestStatus":{"code":100,"result":true},"requestType":"CreateProfile"},"op":7}

Which is abnormal/not the right response I believe? As thats the "it worked" response right?

BarryCarlyon:
### Description

Fixes #1264 

~~This just updates the function name invoked to the new one outlined in https://github.com/obsproject/obs-studio/commit/b9808eaca82340132c56e1577ae58a1e8747a119~~

~~Conceptually to prevent this fault in future should move to `obs_frontend_create_profile` instead?~~

Move create and delete profile to their API functions instead

### Motivation and Context

WebSocket function for `CreateProfile` is not working in 31 beta due to internal function rename and not using an API function

### How Has This Been Tested?

Tested OS(s): 

### Checklist:

-  [ ] The code has been tested.

### DIFF
--- a/src/requesthandler/RequestHandler_Config.cpp
+++ b/src/requesthandler/RequestHandler_Config.cpp
@@ -290,7 +290,7 @@ RequestResult RequestHandler::CreateProfile(const Request &request)
 		return RequestResult::Error(RequestStatus::ResourceAlreadyExists);
 
 	QMainWindow *mainWindow = static_cast<QMainWindow *>(obs_frontend_get_main_window());
-	QMetaObject::invokeMethod(mainWindow, "NewProfile", Qt::BlockingQueuedConnection,
+	QMetaObject::invokeMethod(mainWindow, "CreateNewProfile", Qt::BlockingQueuedConnection,
 				  Q_ARG(QString, QString::fromStdString(profileName)));
 
 	return RequestResult::Success();
--- a/src/requesthandler/RequestHandler_Config.cpp
+++ b/src/requesthandler/RequestHandler_Config.cpp
@@ -289,7 +289,7 @@ RequestResult RequestHandler::CreateProfile(const Request &request)
 	if (std::find(profiles.begin(), profiles.end(), profileName) != profiles.end())
 		return RequestResult::Error(RequestStatus::ResourceAlreadyExists);
 
-	obs_frontend_create_profile(profileName);
+	obs_frontend_create_profile(profileName.c_str());
 
 	return RequestResult::Success();
 }
@@ -322,7 +322,7 @@ RequestResult RequestHandler::RemoveProfile(const Request &request)
 	if (profiles.size() < 2)
 		return RequestResult::Error(RequestStatus::NotEnoughResources);
 
-	obs_frontend_delete_profile(profileName);
+	obs_frontend_delete_profile(profileName.c_str());
 
 	return RequestResult::Success();
 }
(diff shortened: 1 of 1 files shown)


src/requesthandler/RequestHandler_Config.cpp
```diff
@@ -290,7 +290,7 @@ RequestResult RequestHandler::CreateProfile(const Request &request)
 		return RequestResult::Error(RequestStatus::ResourceAlreadyExists);
 
 	QMainWindow *mainWindow = static_cast<QMainWindow *>(obs_frontend_get_main_window());
-	QMetaObject::invokeMethod(mainWindow, "NewProfile", Qt::BlockingQueuedConnection,
+	QMetaObject::invokeMethod(mainWindow, "CreateNewProfile", Qt::BlockingQueuedConnection,
```
PatTheMav, line 293: Plugins are not supposed to invoke private API methods directly - the method names can change at any point (as just happened) and the existence of these functions is not something plugins should know about.

Thus this isn’t a ‘fix’, but instead perpetuates the same mistake that lead to the bug in the first place.

As pointed out in the OP, this should use the front end API instead, which was updated to call the correct (updated) function internally and thus provides the stable interface for plugins.

BarryCarlyon: It's outside my time and ability to file that PR but a simple function name swap was simple enough.

A competiting PR is needed doing it "properly"

BarryCarlyon: And theres a possibly other functions across the websocket plugin (such as `DeleteProfile` that "do it wrong" too, seems a further/in depth audit and revise needed?)

BarryCarlyon: I think that should do it?

src/requesthandler/RequestHandler_Config.cpp
```diff
@@ -289,9 +289,7 @@ RequestResult RequestHandler::CreateProfile(const Request &request)
 	if (std::find(profiles.begin(), profiles.end(), profileName) != profiles.end())
 		return RequestResult::Error(RequestStatus::ResourceAlreadyExists);
 
-	QMainWindow *mainWindow = static_cast<QMainWindow *>(obs_frontend_get_main_window());
-	QMetaObject::invokeMethod(mainWindow, "NewProfile", Qt::BlockingQueuedConnection,
-				  Q_ARG(QString, QString::fromStdString(profileName)));
+	obs_frontend_create_profile(QString::fromStdString(profileName));
```
PatTheMav, line 292: Last I checked, `obs_frontend_create_profile` takes a `const char*`, so this should probably say `profileName.c_str()` instead (the signature is `void obs_frontend_create_profile(const char *name)`).

Granted, _inside_ the API it will create a new `QString` with that, but those are internals that the caller should not care about. There is no need to use _any_ Qt types here, as from the POV of the plugin Qt might as well not exist: There is a frontend, it has an API, and you don't know whether it's Qt, GTK, SwiftUI or WinUI.

PatTheMav (approved):

RytoEX (approved):

RytoEX:
In hindsight, these commits should have been squashed.

Issue #907, opened 2022-02-08
Security Issue: Allow changing of binding options to decrease attack vectors


mitchcapper:
### Operating System Info

Other

### OBS Studio Version

Other

### obs-websocket Version

Git

### OBS Studio Log URL

n/a

### Expected Behavior

n/a

### Current Behavior

n/a

### Steps to Reproduce

/na

### Anything else we should know?

Before the details I figured start with why this matters, what is the attack surface this plugin exposes and what is at risk if a malicious user (you below) can successfully execute commands:

- You can create new scenes/captures of the users entire desktop for example, and then start a remote stream or save the recording remotely. 
- You can write files to the local disk at arbitrary paths you specify (save screenshot for example)
- You can remotely listen to any audio input (microphone) or output on the users pc
- Trigger absolutely any OBS hot key the user has setup
- Read the users configuration for scenes/sources/profiles/stream services/etc (not sure if this could have sensitive information embedded but it looks like it could contain keys)
- Read images/videos from anywhere on the users computer.  You do not need to know the name of the image/video you want to see just using the slideshow functionality you can read entire directories of images/videos.  For example trying common paths for storage/sync services like google drive, dropbox, onedrive, or just the users Pictures and Videos folder in their private account.
- Read other files from the computer, I have no idea on OBS's full capabilities but I at least see a "Text Source" option which I believe can easily access any UTF-8 encoded file.
- I don't see an obvious 'execute random OS command or application' but OBS supports such a massive source of inputs and outputs I might be missing the easy method of doing this.
- OBS does not run in a lower security context, a sandbox, or drop the user's credential tokens (I believe) so likely anything the user has network access to is also fair game (ie a user on a corporate network having all file share access available).
- The fact: All code contains bugs, some bugs can lead to remote code execution / full system access,  unmanaged languages like C/C++ (what this plugin is written in) are more likely to be subject to corruption/off by one exploits.  You have rolled your own security / web socket authentication method (sadly a downside of there not being a standard), meaning less eyes are probably on that code than more standardized code.  You have to allow the user to connect and issue certain commands before you get to authentication, so there is already more potential vulnerability (and parsing of arbitrary json or MsgPack binary data that the client sends) than say a TLS client cert auth stack.

[...]

PR #1279, opened 2025-01-22
websocketserver: Allow the user to bind to a specific address 

silasary:
### Description
This PR adds a config variable and command line argument for binding the websocket server to a specific address, rather than blindly listening on all interfaces.
This PR does not currently add a GUI element to configure this.

By default, OBS will continue listening on all addresses, this just adds the option to reduce the exposure for users who want or need it (changing the default appears to be the main reason 1cd12c10237a388288250820227d05800f40f064 was reverted).

I have chosen to allow binding of a specific address instead of a "localhost only" flag, as it's more flexible for users with multiple network adaptors.

No UI changes just yet, because I'm not sure how to best expose it in a UI, and feel like the users who want/need it are probably comfortable with hand-editing the config or using the CLI argument.  

### Motivation and Context

I am trying to use OSB websockets in an environment where listening on all addresses is a problem.  This allows me to bind OBS to only one network adaptor.
This PR also addresses the security concerns of #907, allowing the user to bind to `127.0.0.1` or `::1`.

### How Has This Been Tested?

Modified config file, tried connecting to the socket from local machine and remote machine.
Reset config file, then repeated test with command line argument.
Also conducted test without any config changes to ensure default was still `0.0.0.0`

Tested OS(s): 
Windows 10, Windows 11

### Types of changes

### Checklist:

-  [ ] I have included updates to all appropriate documentation.

### DIFF
--- a/src/Config.cpp
+++ b/src/Config.cpp
@@ -38,11 +38,13 @@ with this program. If not, see <https://www.gnu.org/licenses/>
 #define PARAM_FIRSTLOAD "first_load"
 #define PARAM_ENABLED "server_enabled"
 #define PARAM_PORT "server_port"
+#define PARAM_HOST "server_host"
 #define PARAM_ALERTS "alerts_enabled"
 #define PARAM_AUTHREQUIRED "auth_required"
 #define PARAM_PASSWORD "server_password"
 
 #define CMDLINE_WEBSOCKET_PORT "websocket_port"
+#define CMDLINE_WEBSOCKET_HOST "websocket_host"
 #define CMDLINE_WEBSOCKET_IPV4_ONLY "websocket_ipv4_only"
 #define CMDLINE_WEBSOCKET_PASSWORD "websocket_password"
 #define CMDLINE_WEBSOCKET_DEBUG "websocket_debug"
@@ -72,6 +74,8 @@ void Config::Load(json config)
 		AuthRequired = config[PARAM_AUTHREQUIRED];
 	if (config.contains(PARAM_PASSWORD) && config[PARAM_PASSWORD].is_string())
 		ServerPassword = config[PARAM_PASSWORD];
+	if (config.contains(PARAM_HOST) && config[PARAM_HOST].is_string())
+		ServerHost = config[PARAM_HOST];
 
 	// Set server password and save it to the config before processing overrides,
 	// so that there is always a true configured password regardless of if
@@ -104,6 +108,17 @@ void Config::Load(json config)
 			blog(LOG_WARNING, "[Config::Load] Not overriding WebSocket port since integer conversion failed.");
 		}
 	}
+
+	// Process `--websocket_host` override
+	QString host_argument = Utils::Platform::GetCommandLineArgument(CMDLINE_WEBSOCKET_HOST);
+	if (host_argument != "") {
+		blog(LOG_INFO, "[Config::Load] --websocket_host passed. Overriding WebSocket host with: %s",
+		     host_argument.toStdString().c_str());
+		ServerHost = host_argument.toStdString();
+		HostOverridden = true;
+	}
+
+
 
 	// Process `--websocket_ipv4_only` override
 	if (Utils::Platform::GetCommandLineFlagSet(CMDLINE_WEBSOCKET_IPV4_ONLY)) {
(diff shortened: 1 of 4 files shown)


src/Config.cpp
```diff
@@ -38,11 +38,13 @@ with this program. If not, see <https://www.gnu.org/licenses/>
 #define PARAM_FIRSTLOAD "first_load"
 #define PARAM_ENABLED "server_enabled"
 #define PARAM_PORT "server_port"
+#define PARAM_HOST "server_host"
 #define PARAM_ALERTS "alerts_enabled"
 #define PARAM_AUTHREQUIRED "auth_required"
 #define PARAM_PASSWORD "server_password"
 
 #define CMDLINE_WEBSOCKET_PORT "websocket_port"
+#define CMDLINE_WEBSOCKET_HOST "websocket_host"
```
tt2468, line 47: I think that I would like this to use `address` instead of `host`. `address` would likely be less ambiguous.

src/Config.cpp
```diff
@@ -72,6 +74,8 @@ void Config::Load(json config)
 		AuthRequired = config[PARAM_AUTHREQUIRED];
 	if (config.contains(PARAM_PASSWORD) && config[PARAM_PASSWORD].is_string())
 		ServerPassword = config[PARAM_PASSWORD];
+	if (config.contains(PARAM_HOST) && config[PARAM_HOST].is_string())
```
tt2468, line 77: Nitpick: Move this and other occurrences above the port override

src/websocketserver/WebSocketServer.cpp
```diff
@@ -100,7 +100,10 @@ void WebSocketServer::Start()
 	_server.reset();
 
 	websocketpp::lib::error_code errorCode;
-	if (conf->Ipv4Only) {
+	if (conf->ServerHost != "") {
+		blog(LOG_INFO, "[WebSocketServer::Start] Locked to %s", conf->ServerHost);
+		_server.listen(conf->ServerHost, std::to_string(conf->ServerPort), errorCode);
```
tt2468, line 105: I'm looking at the websocketpp documentation, and I'm not seeing a matching function for listen(). Do you know which function it's being implicitly cast to? I'm guessing `listen(InternetProtocol const &internet_protocol, uint16_t port, lib::error_code &ec)`

It may be safer in this specific situation to first create an `InternetProtocol` object, then pass it to the listen function.

src/Config.cpp
```diff
@@ -105,6 +109,17 @@ void Config::Load(json config)
...
+		blog(LOG_INFO, "[Config::Load] --websocket_host passed. Overriding WebSocket host with: %s",
+		     host_argument.toStdString().c_str());
+		ServerHost = host_argument.toStdString();
+		HostOverridden = true;
+	}
+
+
+
 	// Process `--websocket_ipv4_only` override
 	if (Utils::Platform::GetCommandLineFlagSet(CMDLINE_WEBSOCKET_IPV4_ONLY)) {
```
tt2468, line 124: This should probably log a warning if the address is overridden, and ensure that the `Ipv4Only` variable is not set to true.

tt2468 (changes requested):

src/websocketserver/WebSocketServer.cpp
silasary: I'm using `listen(	std::string const &host, std::string const &service, lib::error_code &ec)` [docs here](https://docs.websocketpp.org/classwebsocketpp_1_1transport_1_1asio_1_1endpoint.html#a85606665cc9c948194076c2377cb61c0)

As far as I could tell, InternetProtocol doesn't let you bind to a specific address.

src/Config.cpp
silasary: I used host because that's what websocketpp called it, but agreed, address is a better name.

Fenrirthviti:
Giving this a bump, as it has unresolved comments still. Looks like the feedback was acknowledged, but the requested changes have not been made yet.

silasary:
Thanks for the bump 😄

src/websocketserver/WebSocketServer.cpp
RytoEX: This feedback thread is still unresolved. What is the path forward here?

LinuxMainframe:
Giving a bump because I would like this functionality

PR #1314, opened 2025-08-31
requesthandler: Add Canvas support

exeldro:
### Description
Add Canvas support

### Motivation and Context
Scenes in not the main canvas are now not accessible 

### How Has This Been Tested?
using different obs-websocket clients, calling the new `GetCanvasList` and calling the other scene call with and without a canvas
Tested OS(s): Windows 11

### Types of changes

### Checklist:

### DIFF
--- a/src/eventhandler/EventHandler_Canvases.cpp
+++ b/src/eventhandler/EventHandler_Canvases.cpp
@@ -1,0 +1,88 @@
+/*
+obs-websocket
+Copyright (C) 2016-2021 Stephane Lepin <stephane.lepin@gmail.com>
+Copyright (C) 2020-2021 Kyle Manning <tt2468@gmail.com>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along
+with this program. If not, see <https://www.gnu.org/licenses/>
+*/
+
+#include "EventHandler.h"
+
+/**
+ * A new canvas has been created.
+ *
+ * @dataField canvasName | String  | Name of the new canvas
+ * @dataField canvasUuid | String  | UUID of the new canvas
+ *
+ * @eventType CanvasCreated
+ * @eventSubscription Canvases
+ * @complexity 2
+ * @rpcVersion -1
+ * @initialVersion 5.7.0
+ * @api events
+ * @category canvas
+ */
+void EventHandler::HandleCanvasCreated(obs_canvas_t *canvas)
+{
+	json eventData;
+	eventData["canvasName"] = obs_canvas_get_name(canvas);
+	eventData["canvasUuid"] = obs_canvas_get_uuid(canvas);
+	BroadcastEvent(EventSubscription::Canvases, "CanvasCreated", eventData);
+}
+
+/**
+ * A canvas has been removed.
+ *
+ * @dataField canvasName | String  | Name of the removed canvas
+ * @dataField canvasUuid | String  | UUID of the removed canvas
+ *
+ * @eventType CanvasRemoved
+ * @eventSubscription Canvases
+ * @complexity 2
+ * @rpcVersion -1
+ * @initialVersion 5.7.0
+ * @api events
+ * @category canvas
+ */
+void EventHandler::HandleCanvasRemoved(obs_canvas_t *canvas)
+{
+	json eventData;
+	eventData["canvasName"] = obs_canvas_get_name(canvas);
+	eventData["canvasUuid"] = obs_canvas_get_uuid(canvas);
+	BroadcastEvent(EventSubscription::Canvases, "CanvasRemoved", eventData);
+}
+
+/**
+ * The name of a canvas has changed.
+ *
+ * @dataField canvasUuid    | String | UUID of the canvas
+ * @dataField oldCanvasName | String | Old name of the canvas
+ * @dataField canvasName    | String | New name of the canvas
+ *
+ * @eventType CanvasNameChanged
+ * @eventSubscription Canvases
+ * @complexity 2
+ * @rpcVersion -1
+ * @initialVersion 5.7.0
+ * @api events
+ * @category canvas
+ */
+void EventHandler::HandleCanvasNameChanged(obs_canvas_t *canvas, std::string oldCanvasName, std::string canvasName)
+{
+	json eventData;
+	eventData["canvasUuid"] = obs_canvas_get_uuid(canvas);
+	eventData["oldCanvasName"] = oldCanvasName;
+	eventData["canvasName"] = canvasName;
+	BroadcastEvent(EventSubscription::Canvases, "CanvasNameChanged", eventData);
+}
--- a/src/requesthandler/RequestHandler_Canvases.cpp
+++ b/src/requesthandler/RequestHandler_Canvases.cpp
@@ -1,0 +1,62 @@
+/*
+obs-websocket
+Copyright (C) 2016-2021 Stephane Lepin <stephane.lepin@gmail.com>
+Copyright (C) 2020-2021 Kyle Manning <tt2468@gmail.com>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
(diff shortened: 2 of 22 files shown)


nyewkey:
> Canvases (at least per their original concept) are a data structure to allow OBS Studio as well as plugins to generate and access "outputs" that exist in the application. I'd be interested to learn what the use case for a websocket-based client would be to get direct access to this internal data object?
> 
> Conceptually a remote control client (what websockets is made for) should not futz around with canvases - it can interact with OBS (which then uses canvases to achieve the required result) or possibly with a plugin (which again might use a canvas to do so), but the fact that canvases exist is not something an external API should even know about.

This change unintentionally breaks workflows for thousands of users relying on tools like Streamer.bot, MixItUp and the Aitum Vertical plugin.

Without Canvas support, external apps can no longer detect or control sources/scenes in vertical canvases; which were fully functional prior to 32.x.

Re-enabling access (even just via an optional canvas parameter) would restore compatibility without exposing internal structures unnecessarily.

**This regression effectively breaks compatibility with popular automations and streaming workflows that depend on OBS websocket control.**

PatTheMav:
Small addendum: The main problem is one of disambiguating canvases just based on names, which can potentially be done by a change in `libobs` itself that would require no further changes to API consumers. It's not great, but a good-enough compromise and would just be a necessary change in `libobs` before this PR could potentially be merged as-is.

jdudetv:
bumping as this would be such a great feature with stuff like the Aitum Stream Suite adding unlimited canvas generation. Being able to get information about those items would be extremely helpful.

src/requesthandler/RequestHandler_Filters.cpp
```diff
@@ -107,6 +109,8 @@ RequestResult RequestHandler::GetSourceFilterDefaultSettings(const Request &requ
 /**
  * Creates a new filter, adding it to the specified source.
  *
+ * @requestField ?canvasName     | String | Name of the canvas the source is in
+ * @requestField ?canvasUuid     | String | UUID of the canvas the source is in
```
Warchamp7: What's the reason for this addition?

Warchamp7 (commented):
I have very mixed feelings about adding optional canvasName/canvasUuid parameters to a bunch of existing methods simply for handling Scene and Group sources when doing requests by object names rather than UUIDs, but I'm not sure there is a better short term solution.

What is the behaviour of these changes if a request is made for a Source such as `GetSourceFilterList` when `canvasName`/`canvasUuid` are provided but not needed? Is it completely ignored, or is there an unnecessary lookup/validation still?

If the lookup always happens no matter what, what happens when `sourceUuid` is provided and is enough to get the relevant object (Ex. Not a Scene or Group source) but an unnecessary `canvasName`/`canvasUuid` is still provided which is incorrect for that source? Does it serve as an additional way to ensure the correct source and cause an error due to the mismatch or is it ignored?

Warchamp7:
> This change unintentionally breaks workflows for thousands of users relying on tools like Streamer.bot, MixItUp and the Aitum Vertical plugin.

Can you explain how this affects Streamer.bot and MixItUp specifically? Or do you mean their functionality that worked with Aitum Vertical?

protected22:
> > This change unintentionally breaks workflows for thousands of users relying on tools like Streamer.bot, MixItUp and the Aitum Vertical plugin.
> 
> Can you explain how this affects Streamer.bot and MixItUp specifically? Or do you mean their functionality that worked with Aitum Vertical?

This is with more external program's (Touchportal, Streamdeck, Streaminterakt etc) when the person uses the Aitum Vertical plugin in OBS. All external program's don't recognise vertical scenes anymore since OBS version 32 and Aitum Vertical 1.6.0. So all integrations or actions which you can set up in those program's lost all functionality since.

huds601:
Not going to lie, as an alpha and now beta tester for vertical on Twitch, lack of websocket support for the vertical canvas since Aitum switched to the first-party canvas feature has been excrutiatingly painful to deal with. 

Following the developments here and seeing ZERO progress in almost SIX months has been even more frustrating to watch, especially knowing there is a solution offered (albeit maybe not perfect, but a solution nevertheless) yet arguments continue over the specifics of implementation that, frankly, really don't matter compared to the **COMPLETE LOSS OF FUNCTIONALITY**.

There is a clear need for multiple canvas support in obs-websocket for the growing number of vertical streamers. We are in danger of Twitch releasing vertical to all and there still won't be proper support in obs-websocket.

I'm now having to look at asking a developer to code a custom script for Firebot to hook into Aitum's vendor requests. This is likely going to be a stop-gap at best offering minimal support for some of my overlay automation. Worse still, I'm going to end up putting in tens of hours of work updating my automation knowing I'm going to have to eventually redo all of of it when _something_ eventually ships on the websocket side.

**Can you please at least ship SOMETHING to support the growing number of vertical streamers controlling scenes/sources via 3rd party bots using obs-websocket, and worry about the minutia and possible future use-cases later?!**

dennisrijsdijk:
@Warchamp7 has requested testing in discord, here are my findings so far:

I can't get any scene or source related events to fire from alternate canvases. @tt2468 had some comments about this in discord

Nitpick: I don't see an explicit way to know which canvas is the OBS default canvas in GetCanvasList (this is the canvas I would like to present to the user by default when creating new actions)

My use case for knowing which canvas is OBS' default canvas is to pre-fill it in the dropdown, rather than having my own default value, alongside the OBS main canvas. I realize I could implicitly grab it by the name "Main" or using the first canvas returned by array, but unless either of those are recommended as official solutions, I think having an explicit way to do it would be best.

Other than those, I haven't ran into any issues in my testing.

dennisrijsdijk:
I missed CurrentProgramSceneChanged in my previous comments, it's also not firing on an alternate canvas. Though there should probably be an explicitly new event for scene changes on canvases (including main)

SceneNameChanged is still not firing on alternative canvases (could this be related to the plugin which provides the canvas instead of OBS? I do see code changes related to it in this PR)

SceneItemEnableStateChanged still isn't passing along canvasName/canvasUuid

The issue with SetCurrentProgramScene putting the vertical scene on the main OBS canvas is still happening. Though just like CurrentProgramSceneChanged, this may need a new request to change a canvas scene, rather than expanding program scene to handle multiple canvases.

src/eventhandler/EventHandler_Canvases.cpp
```diff
@@ -0,0 +1,88 @@
...
+ * @dataField canvasName | String  | Name of the removed canvas
+ * @dataField canvasUuid | String  | UUID of the removed canvas
+ *
+ * @eventType CanvasRemoved
+ * @eventSubscription Canvases
+ * @complexity 2
+ * @rpcVersion -1
+ * @initialVersion 5.7.0
+ * @api events
+ * @category canvases
```
Warchamp7: Adjust to `@category canvas`

src/requesthandler/RequestHandler_Canvases.cpp
```diff
@@ -0,0 +1,61 @@
...
+ * Gets an array of canvases in OBS.
+ * 
+ * @responseField canvases                  | Array<Object> | Array of canvases
+ *
+ * @requestType GetCanvasList
+ * @complexity 2
+ * @rpcVersion -1
+ * @initialVersion 5.0.0
+ * @api requests
+ * @category scenes
```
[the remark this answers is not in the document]
Warchamp7: Adjust to `@category canvas`

Warchamp7 (changes requested):
`Canvas` will need to be added to the category list in `docs/docs/generate_md.py`.

The places noting `@category canvases` should be adjusted to `@category canvas`

Warchamp7:
> Events without canvasName/canvasUuid
> 
> ~SceneItemCreated SceneItemRemoved SceneItemListReindexed~ SceneItemEnableStateChanged
> 
> ~SceneTransitionStarted CurrentProgramSceneChanged SceneTransitionEnded~
> 
> This is not an exhaustive list of events as I don't subscribe to all events available through the websocket.

I'm not sure if it makes sense for Scene Item events to include Canvas info. They already include the Scene context and the Canvas can be determined using that.

Warchamp7:
> Following the developments here and seeing ZERO progress in almost SIX months has been even more frustrating to watch, especially knowing there is a solution offered (albeit maybe not perfect, but a solution nevertheless) yet arguments continue over the specifics of implementation that, frankly, really don't matter compared to the **COMPLETE LOSS OF FUNCTIONALITY**.

I understand the frustration but we've only had a single OBS release since the time this PR was opened. Even in the best case scenario it would have been fixed in December. It is getting attention now to hopefully be fixed for the next release, or a minor update shortly after.


> > I'm not sure if it makes sense for Scene Item events to include Canvas info. They already include the Scene context and the Canvas can be determined using that.
> 
> unfortunately, I've got legacy setups going back years where the user actions only have the scene name and scene item id, not the scene uuid, and as I recall, scene and group names are not unique across canvases. Having the canvas info included rules out any ambiguity between same-named scenes on different canvases.

At some point in the future the use of name is going to be removed entirely in favor of UUIDs since they are, as the name suggests, unique. Any current setup would fall apart if a scene had two reference copies of the same source right now, as name will grab the first match and there can be multiple.

If you're able to update those setups to utilize the canvas aware data being included by this update, I'd highly recommend  some form of user data update/migration on initialization that updates them to use UUIDs for both handling that edge case as well as future proofing.

PatTheMav (approved):

Warchamp7 (approved):

LinuxMainframe:
Thanks, this fixed my issue @exeldro

PR #1324, opened 2026-02-04
Add "custom type" defintions for protocol.json/protocol.md generation

sid2934:
### Description

This Pull Request hopes to propose a solution to an issue I was facing. Currently the protocol.json/md do not provided full type information for generating clients that call the OBS WebSocket server. Internally obs-websocket is build responses dynamically for a number of request/response properties. This means its flexible, but at the cost of usability when the API is fairly stable at this point.

### Motivation and Context

The motivation here is to introduce light weight mechanism that will allow generators to create fully statically typed models for the request/response objects. This is the less invasive of the two options I considered since it does not modify any of the obs-websocket code, as all of the changes are in comments and the documentation generators only. 

The alternative option is create the needed request/response property type models (likely header files) and enforce the use of statically typed objects throughout the code base. These new models work require the same/similar annotation to produce the proper protocol.json.

### How Has This Been Tested?

Generators are working, no core changes to obs-websocket functionality

Tested OS(s): 
Windows and MacOS 

### Types of changes

### Checklist:


sid2934:
@BarryCarlyon, @DDRBoxman, @RytoEX, @Rosuav

Bumping this PR and discussion to see if I can get traction for adding the custom type definitions to the protocol files.

